PR #37, opened 2014-11-18
Support unicode characters

anler:
Hi, I was trying to draw some unicode characters (lambdas) and it was failing, this is a fix for that (not taking python3 compatibility into account)

thanks


hagberg:
I'm not sure what the problem is here.  But this isn't likely the solution.  If there is a specific issue please reopen with an example.

PR #45, opened 2015-02-22
Update INSTALL.txt

xhudik:
In INSTALL.txt:
"(optional) Run "python setup_egg.py nosetests" to execute the tests"
1. there is no target "nosetests". 
2. running python setup_egg.py test gives strange:
setup_egg.py: error: Error reading config file 'setup.cfg': no such option 'with-doctest-ignore-unicode'
Therefore I suppose setup_egg.py nosetests is obsolete (not-functional) and should be replaced by standard setup.py test

I'm not completely sure - pls. double check whether my hypothesis is correct

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -47,7 +47,7 @@ Source Archive File
 
   3. Run "python setup.py install" to build and install 
 
-  4. (optional) Run "python setup_egg.py nosetests" to execute the tests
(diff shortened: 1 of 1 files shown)


hagberg:
These are optional tests,  and they do work correctly.
You need nose, and doctest-ignore-unicode

e.g.
$ pip install nose
$ pip install doctest-ignore-unicode

Issue #10, opened 2014-02-20
pip install with --install-option


j0hnsmith:
It would be great if it was possible to install pygraphviz via pip and pass `library_path` and `include_path` arguments eg 

    pip install pygraphviz --install-option="--library_path=/my/lib/path" --install-option="--include_path=/my/include/path"

hagberg:
Any update here?
This is the only issue open for the next release.

PR #52, opened 2015-05-19
Implement library path

mmarkov:
I reworked a little bit installation routine and moved everything to setuptools.  Also I implemented ability to use install options #10


hagberg:
Looks good! I'm going to merge this so we can get some feedback from the community.

PR #64, opened 2015-08-13
PyPy support

jamadden:
This PR brings support for PyPy 2.5.0 and above (tested with 2.5.0 and 2.6.0 on Linux and 2.5, 2.6, and 2.7 on OS X). All the tests pass on Travis.

The major change is in `graphviz.i`'s unwrapping of FILE* objects. On PyPy, the `PyFile_AsFile` function is not supported, so we go with the codepath that Python 3 uses. This of course required the .c file to get rebuilt, and I used the most current version of SWIG available to do so---which complained about syntax in the copyright header lines ("Error: Unknown SWIG preprocessor directive: Copyright", etc). I wasn't sure what to do there; hopefully a comment is ok.

(There's also some automatic whitespace cleanup in the .i file; emacs does that automatically and I forgot to turn that off, but I can restore the whitespace if necessary to make the diff cleaner.)

The other notable change has to do with a build-time issue. I couldn't figure out how to get PyPy on Travis (or a local Linux) to properly link and load the graphviz libraries with the PyPy module, so I took the blunt approach with LD_PRELOAD. I then had to undo this hammer in agraph.py. I'd like to have a better way, but at least this gets things working.

I'm happy to make any adjustments needed. This is my first time contributing to this project.


hagberg:
Thanks for looking into this.   It seems pretty straightforward to add pypy support this way.  It would be good to figure out how to do the build/link correctly.

jamadden:
I totally agree, but I'm just about out of ideas. It works perfectly on OS X, there's only a problem on Linux. The complier args look correct there, though. I'm guessing its something to do with "as needed" link removal, something that maybe differs between gcc and clang, but I haven't so far come up with what to do about it. Any ideas are welcome!

At any rate, this is a blocker for us and probably others using pypy in production, so at least this much works :)

jamadden:
It turned out that simply putting the libraries in their dependency order in `setup.py` fixed the issues under GCC/GNU ld. No more LD_PRELOAD!

jamadden:
If there's anything I can do to facilitate this, please let me know.

jamadden:
Closing in favor of #75.

Issue #65, opened 2015-08-16
py3k pygraphviz cant generate simple graphs


sandrotosi:
Hello,
as reported at http://bugs.debian.org/794444 it seems py3k pygraphviz is unable to generate even simple graphs, for example use:

python3 -c 'import networkx as nx; G=nx.Graph();G.add_node(1,foo="bar");nx.write_dot(G, "out")'

hagberg:
Here is a simpler version that triggers the bug

python3 -c "import pygraphviz as pgv; A = pgv.AGraph(); A.add_node(1,foo='bar');A.clear()"

(The pygraphviz.AGraph.clear() method is called in nx.write_dot(),  removing that resolves this problem in networkx but will cause memory leaks)

I haven't been able to track down the exact problem.  It is triggered by clearing the Attribute() class items in the call to AGraph.clear() on any of the three lines 

```python
        self.edge_attr.clear()
        self.node_attr.clear()
        self.graph_attr.clear()
```

The Attribute class is a DictMixin subclass that manipulates attributes using the libcgraph API.
It's possible some of the wrappers used there, generated by SWIG to the libcgraph API,  aren't working as expected.  The bad behavior is a hang in the loop of Attribute.iteritems() so perhaps agnxtattr isn't working correctly with the Python3 wrappers?


I'd be grateful for help debugging this.

mmarkov:
@hagberg, you are absolutely right. It looks like Attribute.__delitem__ doesn't delete item, so it never throws KeyError and hang up in infinite loop.

hagberg:
I couldn't figure out how to fix it when I looked at it.  Any idea?

mmarkov:
I don't promise, but I will try to take a look at this over next few
days. It doesn't look complex and I already has unit test which reproduce
error.

Max

hagberg:
resolved with #67 and #68

PR #67, opened 2015-08-31
Implement changes to AGraph.clear() to handle custom attributes

mmarkov:
Fixed issue #65 
Actually node_attr is dictionary of all possible node attributes and it should not be cleared during graph clear. It will be cleared automatically during handle close.


hagberg:
Looks good. Thanks.

mmarkov:
Hi  Aric,

I found one more issue and fixed it in one more pull request.

with kind regards,
Max

Issue #72, opened 2015-09-10
Import error on Ubuntu 15.04


Qwlouse:
I've installed pygraphviz 1.3.1 with pip on Ubuntu 15.04, and when I try to import I get the following error:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-99ed911275c0> in <module>()
----> 1 import pygraphviz

/home/greff/venv/py2/local/lib/python2.7/site-packages/pygraphviz/__init__.py in <module>()
     56 __license__  = release.license
     57 
---> 58 from .agraph import AGraph, Node, Edge, Attribute, ItemAttribute, DotError
     59 
     60 __all__=[

/home/greff/venv/py2/local/lib/python2.7/site-packages/pygraphviz/agraph.py in <module>()
     24     from collections import MutableMapping as DictMixin
     25 
---> 26 from . import graphviz as gv
     27 
     28 _DEFAULT_ENCODING = 'UTF-8'

/home/greff/venv/py2/local/lib/python2.7/site-packages/pygraphviz/graphviz.py in <module>()
     26                 fp.close()
     27             return _mod
---> 28     _graphviz = swig_import_helper()
     29     del swig_import_helper
     30 else:

/home/greff/venv/py2/local/lib/python2.7/site-packages/pygraphviz/graphviz.py in swig_import_helper()
     22         if fp is not None:
     23             try:
---> 24                 _mod = imp.load_module('_graphviz', fp, pathname, description)
     25             finally:
     26                 fp.close()

ImportError: /home/greff/venv/py2/local/lib/python2.7/site-packages/pygraphviz/_graphviz.so: undefined symbol: Agundirected
```
Any ideas?

mmarkov:
It looks like same as #71. Try installing with direct path to graphviz
```
pip install pygraphviz --install-option="--include-path=/usr/include/graphviz" --install-option="--library-path=/usr/lib/graphviz/"
```

Qwlouse:
That did the trick. Thank you!

Issue #71, opened 2015-09-08
On Ubuntu 12.04: _graphviz.so: undefined symbol: Agundirected


elistevens:
I've done the following on my dev hosts for a while now. I needed to bring up pygraphviz on a somewhat-freshly imaged host, and got a new error that I've not seen before when I tried to import (see last output; install seems to be smooth):

$ dpkg -s graphviz-dev
Package: graphviz-dev
Status: install ok installed
Priority: optional
Section: devel
Installed-Size: 90
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: all
Source: graphviz
Version: 2.26.3-10ubuntu1.2
Depends: libgraphviz-dev

v$ pip install pygraphviz
Downloading/unpacking pygraphviz
  Downloading pygraphviz-1.3.1.tar.gz (103kB): 103kB downloaded
  Running setup.py (path:/home/elis/venv/mobius3d_dev/build/pygraphviz/setup.py) egg_info for package pygraphviz
    
    warning: no previously-included files matching '*~' found anywhere in distribution
    warning: no previously-included files matching '*.pyc' found anywhere in distribution
    warning: no previously-included files matching '.svn' found anywhere in distribution
    no previously-included directories found matching 'doc/build'
Installing collected packages: pygraphviz
  Running setup.py install for pygraphviz
    Trying pkg-config
    include_dirs=/usr/include/graphviz
    library_dirs=
    
    warning: no previously-included files matching '*~' found anywhere in distribution
    warning: no previously-included files matching '*.pyc' found anywhere in distribution
    warning: no previously-included files matching '.svn' found anywhere in distribution
    no previously-included directories found matching 'doc/build'
    building 'pygraphviz._graphviz' extension
    gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/include/graphviz -I/usr/include/python2.7 -c pygraphviz/graphviz_wrap.c -o build/temp.linux-x86_64-2.7/pygraphviz/graphviz_wrap.o
    pygraphviz/graphviz_wrap.c: In function ‘agattr_label’:
    pygraphviz/graphviz_wrap.c:3321:5: warning: return makes integer from pointer without a cast [enabled by default]
    pygraphviz/graphviz_wrap.c: In function ‘_wrap_agread’:
[...]

PR #73, opened 2015-09-11
Auto detect graphviz installation path on ubuntu

mmarkov:
- implemented search for graphviz installation using dpkg
- improve handling of empty strings, now if path not found error message is generated
- added unit tests to cover new functionality
- this pull request addresses installation issues for Ubuntu (reported in #72  and #71 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ before_install:
   # We need GraphViz
   - "sudo apt-get install graphviz 2>&1 | tail -n 2"
 install:
+  - pip install mock
   - pip install nose
   - pip install doctest-ignore-unicode
   - pip install coveralls
--- a/pygraphviz/tests/test_setup.py
+++ b/pygraphviz/tests/test_setup.py
@@ -2,7 +2,7 @@ import sys
 sys.path.insert(0,'../..')
 import setup_extra
 from nose.tools import *
-from unittest.mock import patch
+from mock import patch
 
 
 def _get_values():
--- a/pygraphviz/tests/test_setup.py
+++ b/pygraphviz/tests/test_setup.py
@@ -1,0 +1,36 @@
+import sys
+sys.path.insert(0,'../..')
+import setup_extra
+from nose.tools import *
+from unittest.mock import patch
+
+
+def _get_values():
+    return None, 'library'
+
+
+def _get_dpkg_output():
+    s = bytes("""libgvc: /usr/local/include/graphviz/gvc.h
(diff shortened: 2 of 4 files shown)


hagberg:
This looks like it should work.  I wonder if there is a simpler solution.  The problem seems to be that the default pkg-config output doesn't include -L/usr/lib - it does if you install into /usr/local/lib, e.g.
$ pkg-config --libs /usr/lib/pkgconfig/libcgraph.pc 
-lcgraph -lcdt
$ pkg-config --libs /usr/local/lib/pkgconfig/libcgraph.pc 
-L/usr/local/lib -lcgraph -lcdt 

Does it solve this problem if we add -L/usr/lib if no path is given?

mmarkov:
@hagberg, I tried install pygraphviz 1.3.1 on clean ubuntu and the challenge is that it installs but doesn't work. As far as I understand to compile pygraphvix SWIG wrapper it is enough to provide only  -lcgraph -lcdt. But as cgraph is abstract library, it doesn't work unless we link with actual implementation (appropriate so or dynlib files). Normally if someone installs graphviz using apt it is installed to /usr/local/lib/graphviz and linker needs this folder to be included into linking path.

Ah, and there is no pkg-config on newly installed ubuntu.

hagberg:
OK - maybe there is more than one issue here?
1) pkg-config is not part of the default ubuntu install - I suggest we require/use pkg-config and note that users need to install this
2) pkg-config doesn't produce -L/usr/lib with the default graphviz install in /usr - maybe that is the way it is supposed to work but pygraphviz doesn't build without that.  

Am I understanding this correctly?

mmarkov:
Yes, you understand it correctly. 
1) Unfortunately, installing pkg-config is not enough - I tried installing it and it doesn't report path to .so files.
2) I tested proposed solution with clean ubuntu and pig-config. Actually -L/usr/lib is not enough because apt installs pygraphviz to -L/usr/lib/pygraphviz and path to this path actually required for proper linking. pygraphviz libraries are not installed in /usr/lib.

j9ac9k:
I can confirm having the same issue, running ubuntu server 15.04 

pip install pygraphviz fails with the following error:

```
  pygraphviz/graphviz_wrap.c:2954:29: fatal error: graphviz/cgraph.h: No such file or directory
   #include "graphviz/cgraph.h"
                               ^
  compilation terminated.
  error: command 'gcc' failed with exit status 1

  ----------------------------------------
  Failed building wheel for pygraphviz
```

I do have pkg-config installed

PR #75, opened 2015-09-16
PyPy support

jamadden:
This is a cleaned up version of the PyPy support from #64, rebased on 1.3.1. The whitespace changes in graphviz.i are gone, but graphviz_wrap.c and graphviz.py were still rebuilt with the latest release of swig.


jamadden:
While testing under OS X 10.11 with graphviz 2.38, I also encountered a reproducible crash having to do with `AGraph.clear()` leaving dangling handles around in `node_attr`, etc. I added a fix for that here.

Travis is green, but coveralls reports a 0.2% decline in coverage. I believe this is in the Swig auto-generated code, a difference due to using a newer Swig. I'm not sure what to do about that or if it matters.

jamadden:
Thank you for the merge! If there can be a Pypi release with these fixes soon, we'd surely appreciate it. Please let me know if I can help in any way.

PR #87, opened 2016-02-24
Fix INSTALL.txt example

rjbez17:
If you use the example as is you have the following:

```
pygraphviz/graphviz_wrap.c:2987:10: fatal error: 'graphviz/cgraph.h' file not found
#include "graphviz/cgraph.h"
```


hagberg:
Most likely missing the full graphviz source or development package which includes cgraph.h

PR #95, opened 2016-06-29
Convert agraph.py warnings from bytes to strings

### LINKED ISSUE #94: Python 3.4 Windows byte decoding error in agraph.py
Python 3.4 causes a byte decoding error on Windows due to agraph.py Warnings being raised from a joined `bytes` object instead of a normal `str` object which has unicode encoding.

The versions of software in use are:

Windows 10 Version 1511 64-bit
Python 3.4.4 32-bit
Graphviz 2.38 32-bit
pygraphviz 1.3.1 32-bit build from http://www.lfd.uci.edu/~gohlke/pythonlibs/
django_extensions 1.6.7

When running `graph_models` command which uses pygraphviz for graph generation with the `django_extensions` package the error is as following:

```
$ ./manage.py graph_models -a -g -o my_project_visualized.png
Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django\core\management\__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django\core\management\__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django\core\management\base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django\core\management\base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django_extensions\management\utils.py", line 58, in inner
    ret = func(self, *args, **kwargs)
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django_extensions\management\commands\graph_models.py", line 162, in handle
    self.render_output_pygraphviz(dotdata, **options)
  File "C:\Users\me\project\virtualenv32\lib\site-packages\django_extensions\management\commands\graph_models.py", line 202, in render_output_pygraphviz
    graph.layout(prog=kwargs['layout'])
  File "C:\Users\me\project\virtualenv32\lib\site-packages\pygraphviz\agraph.py", line 1358, in layout
    data = self._run_prog(prog, ' '.join([args, "-T", fmt]))
  File "C:\Users\me\project\virtualenv32\lib\site-packages\pygraphviz\agraph.py", line 1338, in _run_prog
    warnings.warn(b"".join(errors), RuntimeWarning)
TypeError: can't use a string pattern on a bytes-like object
```

The excepted output depicting font warnings would be on Windows:

```
WARNING:py.warnings:C:\Users\me\project\virtualenv32\lib\site-packages\pygraphviz\agraph.py:1338: RuntimeWarning:
(neato.exe:30444): Pango-WARNING **: couldn't load font "Helvetica Bold Not-Rotated 8", falling back to "Sans Bold Not-Rotated 8", expect ugly output.

(neato.exe:30444): Pango-WARNING **: couldn't load font "Helvetica Italic Not-Rotated 8", falling back to "Sans Italic Not-Rotated 8", expect ugly output.

(neato.exe:30444): Pango-WARNING **: couldn't load font "Helvetica Not-Rotated 8", falling back to "Sans Not-Rotated 8", expect ugly output.

(neato.exe:30444): Pango-WARNING **: couldn't load font "Helvetica ItalicBold Not-Rotated 8", falling back to "Sans Not-Rotated 8", expect ugly output.

(neato.exe:30444): Pango-WARNING **: couldn't load font "Helvetica Bold Not-Rotated 12", falling back to "Sans Bold Not-Rotated 12", expect ugly output.

  warnings.warn(b"".join(errors).decode(), RuntimeWarning)
```

[The crashing segment of code from agraph.py line 1339 is](https://github.com/pygraphviz/pygraphviz/blob/28b9e9d0cefc107f19518a001566b48e1f3c767d/pygraphviz/agraph.py#L1339):

```
        if not data:
            raise IOError(b"".join(errors))

        if len(errors) > 0:
            warnings.warn(b"".join(errors), RuntimeWarning)
```

After making the following fix the program runs normally on my local machine on Python 3 32-bit, 64-bit, and Python 2.7 64-bit:

```
        if not data:
            raise IOError(b"".join(errors).decode())

        if len(errors) > 0:
            warnings.warn(b"".join(errors).decode(), RuntimeWarning)
```

I am proposing that to support Python 3 graph generation a bytes decoding is added to the error-causing code segment in

aleksihakli:
Warning and error raising behaviour is not correct on Python 3,
where bytes objects cannot be implicitly converted to strings,
and the warnings cause a TypeError crash.

Fixes #94

### DIFF
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -1337,10 +1337,10 @@ class AGraph(object):
             t.join()
 
         if not data:
-            raise IOError(b"".join(errors))
+            raise IOError(b"".join(errors).decode(self.encoding))
 
         if len(errors) > 0:
-            warnings.warn(b"".join(errors), RuntimeWarning)
+            warnings.warn(b"".join(errors).decode(self.encoding), RuntimeWarning)
 
         return b"".join(data)
 


hagberg:
That looks like the right fix.  We should figure out how to add a test for this.

aleksihakli:
I don't see an easy way for testing `_run_prog` `RuntimeWarnings` outside of Windows. The specific needed `IOError`s might be possible to test by mocking failed data output, but that is outside my expertise. Hence I can't contribute on the tests here. 

The `.decode` functionality should work on Linux and Windows platforms since it is the same as other `.decode` uses in the same program and file.

I hope we can merge this since currently the program can't be used on Windows on Python 3 and the patch does not regress any functionality.

aleksihakli:
What would you propose as the best way to test this? In the end it's about byte string decoding, which is a fairly standard operation in itself. To have a reasonable test set we would have to cover a wide range of encodings, which can be a difficult thing to test.

I understand the need for automated tests but think that this part that of

hagberg:
Fixed in https://github.com/pygraphviz/pygraphviz/commit/272a113976b809d46d3a1c3a95d00bd82c6c93a8

PR #104, opened 2016-09-30
The warnings module expects a string, not bytes.

jonathanslenders:
Otherwise, it crashes with "TypeError: can't use a string pattern on a
bytes-like object" when the errors list is not empty.


mmarkov:
Looks good, definitely should solve the issue.

Issue #125, opened 2017-05-08
Inexplicable line breaks and character escapes on non-trivial HTML


waynr:
```
zsh/3 2993  (git)-[qeng-4804]-% ipython
Python 3.5.3 (default, Jan 19 2017, 14:11:04)
Type "copyright", "credits" or "license" for more information.

IPython 5.3.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pygraphviz as pgv

In [2]: g = pgv.AGraph()

In [3]: s = '''<<TABLE BORDER="0">
   ...:   <TR>
   ...:       <TD> {display_name} </TD>
   ...:   </TR>
   ...:   <TR>
   ...:       <TD>
   ...:         <TABLE>
   ...:           <TR>
   ...:           <TD align="left">Count</TD>
   ...:           <TD align="right"> {run_count} </TD>
   ...:           </TR>
   ...:           <TR>
   ...:           <TD align="left">Earliest Run</TD>
   ...:           <TD align="right"> {run_earliest} </TD>
   ...:           </TR>
   ...:           <TR>
   ...:           <TD align="left">Latest Run</TD>
   ...:           <TD align="right"> {run_latest} </TD>
   ...:           </TR>
   ...:           <TR>
   ...:           <TD align="left">Avg Runtime</TD>
   ...:           <TD align="right"> {run_time_average} </TD>
   ...:           </TR>
   ...:           <TR>
   ...:           <TD align="left">Avg Failure Rate</TD>
   ...:           <TD align="right"> {failure_rate} </TD>
   ...:           </TR>
   ...:         </TABLE>
   ...:       </TD>
   ...:   </TR>
   ...: </TABLE>>
   ...: '''

In [4]: g.add_node(b, label=s)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-4-5b25408ca4cb> in <module>()
----> 1 g.add_node(b, label=s)

NameError: name 'b' is not defined

In [5]: b = object()

In [6]: g.add_node(b, label=s)

In [7]: print(g)
strict graph {
        node [label="\N"];
        "<object object at 0x7fed8afd0610>"      [label="<<TABLE BORDER=\"0\">
  <TR>
      <TD> {display_name} </TD>
  </TR>
  <TR>
      <TD>
[...]

PR #126, opened 2017-05-08
#125: Add test showing erroneous string mutation in AGraph node label…

waynr:
Initially all this PR does is add a test case demonstrating the string mutation that leads to a borked html formatted label. The test case fails as described in #125.

I'd be happy to hack at this if someone could give me a hint or two as to where this string mutation might be occurring.

### DIFF
--- a/pygraphviz/tests/test_html.py
+++ b/pygraphviz/tests/test_html.py
@@ -3,6 +3,60 @@ from nose.tools import *
 import pygraphviz as pgv
 from os import linesep
 
+long_html_string = '''
+<<TABLE BORDER=0>
+  <TR>
+      <TD> meow </TD>
+  </TR>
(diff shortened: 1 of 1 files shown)


waynr:
Whoops, I made a mistake in the test...gonna fix that shortly.

waynr:
After a brief spelunking into pygraphviz code, I found that the problem is an intolerance of whitespace at either end of of the label string, meaning that the very first character of the label attribute must be a `<` and the very last character of the label attribute must be a `>`. See https://github.com/pygraphviz/pygraphviz/blob/master/pygraphviz/graphviz.i#L213-L254

This should probably be documented somewhere...

hagberg:
Yes, you are correct that the first char has to be < and the last >.

PR #128, opened 2017-06-14
Fix segfault on attribute changes after layout.

pope1ni:
Simple reproducible crash fixed by this commit:

```
>>> import pygraphviz
>>> G = pygraphviz.agraph.AGraph()
>>> G.layout(prog='dot')
>>> G.graph_attr['label'] = 'ABC'
Segmentation fault (core dumped)
```

This is similar to another segmentation fault related to handle changes that was fixed by 3bfd923.


pope1ni:
Test failures seem to be unrelated.

hagberg:
Yes, definitely a bug.  If you can see where the problem is please update this.

pope1ni:
@hagberg: I'm not sure what you mean... This PR does fix the bug...

pope1ni:
Thanks!

PR #129, opened 2017-06-14
Fix tests for graphviz 2.40.1

pope1ni:
It is likely that you'll want to apply these changes conditionally which I haven't done as yet.


hagberg:
I'd like to just leave it break so we can recognize the broken graphviz lib.

PR #146, opened 2018-03-09
Replace raise StopIteration with return

levi-rs:
- Raising StopIteration was deprecated in python 3.5, and throws
   warnings when used
 - Replace `raise StopIteration` with canonical `return`
 - See [PEP 479](https://www.python.org/dev/peps/pep-0479/)


dschult:
superceded by #164 
It turns out we needed to also check for StopIteration being created in the C code. so we can catch it before returning.

Thanks for this!

PR #183, opened 2018-12-31
Fix graphviz_wrap.c for Pypy3

### LINKED ISSUE #177: Segfault when reading a GraphViz file with pypy3
I installed Pypy3 6.0.0 and GraphViz 2.40.1 using Linuxbrew `brew install pypy3 graphviz`, and I installed Pygraphviz 1.5 and Networkx 2.2 using `pip_pypy3 install --user pygraphviz networkx`.

When I attempt to write a GraphViz file using `networkx.write_dot`, it segafults in `_IO_new_fdopen iofdopen.c:63` called by `_wrap_agwrite` in `_graphviz.pypy3-60-x86_64-linux-gnu.so`. It appears to be dereferencing a null pointer. Here's the backtrace:

```
(gdb) bt
#0  _IO_new_fdopen (fd=1, mode=0x0) at iofdopen.c:63
#1  0x00007fc0ef4fb759 in _wrap_agwrite ()
   from /home/sjackman/.local/lib/python3.5/site-packages/pygraphviz/_graphviz.pypy3-60-x86_64-linux-gnu.so
#2  0x00007fc0ed183984 in ?? ()
   from /gsc/btl/linuxbrew/Cellar/pypy3/6.0.0/lib/libpypy3-c.so
(gdb) x/i $pc
=> 0x7fc0eb9a67b <_IO_new_fdopen+11>:	movzbl (%rsi),%eax
(gdb) p/x $rsi
$1 = 0x0
```

This works as expected when using CPython 3.7.0 rather than Pypy3. For my purpose, I have found Pypy3 to be four times faster than CPython. Any thoughts on how to troubleshoot this issue?

khoroshilov:
027d11d ("Fix graphviz_wrap.c _wrap_agwrite for Pypy3") covers _wrap_agwrite,
but leaves the same problem in _wrap_agread because it fixes code in
the file generated by swig.

The commit introduces additional disturbance due using newer swig version:
3.0.8 instead of 3.0.7.

Fixes #177.


dschult:
I believe this change is no longer needed after #230 was merged. no check for version before 3.0 is needed.

PR #189, opened 2019-02-13
WIP: use skbuild to reduce build complexity

Erotemic:
The scikit-build project has a goal of reducing the difficulty running Python setup with compiled extension modules. It does this by acting as a seamless glue between setup.py and CMakeLists.txt. 

Of course all of the complexity then moves to CMakeLists, however that's a good thing because CMake helps to some degree. CMake is inherently cross platform and there is a large ecosystem of package support. However, writing a CMakeLists.txt can still be tricky, which is why this is a WIP.

The reason for this PR is because pygraphviz has always been a pain when it comes to pip installing it. Its easy to pip install opencv and PyQt5 these days, so it should be easy to install pygraphviz too. My hope is that by moving to skbuild it becomes easy to make a robust installable package.


Erotemic:
@dschult This is currently "working on my machine" (tm). I'm currently trying to figure out why its not working on Travis. In the meantime, a code review may be worthwhile.

dschult:
I'm afraid this is beyond my expertise. Not even sure how to test it to see if it works on my machine.
Any hints on good ways to check if it works on a local install?

Erotemic:
I was able to get the build to work in a conda 3.7 environment. I installed the graphviz libraries through conda. I haven't yet been able to make it work with system graphviz libs, but I also didn't look into it too deeply (the error I saw was probably due to a local issue on my system). 

Adding the --verbose to pip in travis just makes the travis error more visible. I haven't figure it out yet, but I'm slowly working on it.

dschult:
PyPy fails during setup because it can't find a ```python_library``` (python_library is None [here](https://github.com/scikit-build/scikit-build/blob/master/skbuild/cmaker.py#L202).  And there is a TODO comment about it [here](https://github.com/scikit-build/scikit-build/blob/master/skbuild/cmaker.py#L422)).  I don't know enough about skbuild or python libraries to see how this code should be changed to work with pypy.  

It looks like skbuild and pypy are incompatible at this time. (Their names and overall goals suggest they will not become compatible... one tries to connect python with C and the other tries to remove python from C :)

dschult:
All non-pypy tests are passing.  The pypy tests fail installing skbuild.

Erotemic:
Did the pypy versions ever build and hook into the C extensions? Very quickly, looking at the code it seems pypy does have support for C-extensions? So I guess this is a failure of scikit-build. 

If this is a scikit-build issue, perhaps we can patch that to include pypy support, in which case this could be triaged until that is ready. Looks like there is an issue here: https://github.com/scikit-build/scikit-build/issues/419

I may look into getting that fixed if I have some time.

dschult:
Yes, the immediate problem is that skbuild does not set a return value for ```python_library``` if one is not found. The resulting value of ```None``` leads to:
```TypeError: unsupported operand type(s) for +: 'str' and 'NoneType'```
I believe the value is supposed to hold a path to the libpython library file. I looked for such a library file for pypy on my system and couldn't find one.  I'm not sure what from pypy would be able to play the role of libpython from CPython.   Sorry for this brain-dump without any real information, but hope it helps.

MridulS:
With the wheels out, we should probably have a look again at using scikit-build-core but lots has changed in 7 years so I'll close the PR for now, but we should get back to this hopefully sometime in the future!

PR #191, opened 2019-04-04
doc: use add_object_type() instead of deprecated add_description_unit()

keszybz:
This is necessary to build the docs with sphinx-2.x.

sphinx-2.0 removed the deprecated function.
See https://bugs.python.org/issue34967 for similar changes in cpython.


dschult:
Thanks!

Issue #137, opened 2017-10-02
self-edges - KeyError: 'agedge: no key'  on  Windows


fpettinari:
On *nix it works, on Windows it raises this error
with Graphviz 2.38.0.1

```
import pygraphviz as pgv
G = pgv.AGraph()
G.add_node('a')
G.add_edge('a', 'a')
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\pygraphviz\agraph.py", line 477, in add_edge
    eh = gv.agedge(self.handle, uh, vh, key, _Action.find)
KeyError: 'agedge: no key'
```

Started with pytransitions/transitions#258

adamasstokhorst:
Posting in this issue again because it's still not resolved, at least on Windows. It's definitely not Graphviz's fault since a manually crafted dot file with self-loops still renders correctly.
```
> python -c "import pygraphviz as pgv; print pgv.__version__"
1.5

> dot -V
dot - graphviz version 2.38.0 (20140413.2041)

> python -c "import sys; print sys.version"
2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:22:17) [MSC v.1500 32 bit (Intel)]

> python -c "import pygraphviz as pgv; G = pgv.AGraph(strict=False); G.add_edge('a',
'a')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Python27\lib\site-packages\pygraphviz\agraph.py", line 479, in add_edge
    eh = gv.agedge(self.handle, uh, vh, key, _Action.find)
KeyError: 'agedge: no key'
```

0x410c:
any update?

jarrodmillman:
We no longer support 2.38.

PR #199, opened 2019-07-09
Update star.py

kevinanewman:
codes fails when i = 0 (A.add_edge(0,0) doesn't work)
range(1, 16) starts i at 1

### DIFF
--- a/examples/star.py
+++ b/examples/star.py
@@ -29,7 +29,7 @@ A.node_attr['fixedsize']='true'
 A.node_attr['fontcolor']='#FFFFFF'
 
 # make a star in shades of red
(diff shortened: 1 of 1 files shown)


kevinanewman:
This is probably related to this: https://github.com/pygraphviz/pygraphviz/issues/137#issue-262044684, but at least this fix should run on any machine.

dschult:
We should merge this even though the example works fine except on GV 2.38 on Windows where self-loops are apparently broken. The star graph doesn't usually include a self loop for the middle node. And it makes it work on the machines/GV versions for which it currently doesn't work.

PR #200, opened 2019-08-04
Add additional include path

adarshp:
Adding an additional include path to find `graphviz/cgraph.h`. For example, running 

```
pkg-config --cflags-only-I libcgraph
```

with MacPorts returns

```
-I/opt/local/include/graphviz
```

(and similarly for Homebrew, except for `usr` instead of `opt`)

### DIFF
--- a/setup_extra.py
+++ b/setup_extra.py
@@ -78,7 +78,10 @@ def _pkg_config():
         output = _b2str(output)
         if output:
             include_path = output.strip()[2:]
-            include_path = include_path.strip() or None
+            include_path = include_path.strip()
+            # This line below adds an extra include path for certain cases where pkg-config 
+            # returns the full path to the cgraph.h directory (e.g. with Homebrew and MacPorts.
+            include_path = include_path + ":" + "/".join(include_path.split("/")[:-1]) or None
     except OSError:
         print("Failed to find pkg-config")
     return include_path, library_path


dschult:
This seems reasonable to me -- though I wonder how many of these we might eventually get.
Is there a way to make pkg-config return what we want?  We could also merge this and then add an issue to explore other ways of finding include paths.

adarshp:
This solution does use pkg-config though - it doesn't add any hard-coded paths.

The command `pkg-config --cflags-only-I libcgraph` returns the following:

- `-I/opt/local/include/graphviz` on macOS with MacPorts
- `-I/usr/include/graphviz` on CentOS and Ubuntu

The line below:

https://github.com/pygraphviz/pygraphviz/blob/f5033cd782c4521d6bc198d5fe85c9ec6938e262/pygraphviz/graphviz_wrap.c#L2676

Seems to work on Linux, but I think (and this is just a guess) that is because it is using `/usr/local/include` as a default path to search for include declarations (in this case, it would find `<graphviz/cgraph.h>` whether or not pkg-config was used). To really utilize the info provided by `pkg-config` and not implicitly rely on the `/usr/local` prefix, I would suggest changing 

```c
#include <graphviz/cgraph.h>
```

to

```c
#include <cgraph.h>
```

PR #212, opened 2020-03-22
Fix graphviz_wrap.c _wrap_agread for Pypy3

nehaljwani:
Similar to https://github.com/pygraphviz/pygraphviz/issues/177 , but for `_wrap_agread`


dschult:
Identical to #183

PR #216, opened 2020-04-08
AGraph Graphviz handle close mechanism

CristiFati:
Browse code while investigating [\[SO\]: Pygraphviz crashes after drawing 170 graphs (@CristiFati's answer)](https://stackoverflow.com/questions/60876623/pygraphviz-crashes-after-drawing-170-graphs/61050313#61050313).

Although closing the handle doesn't have any visible effect, it's better to happen automatically when an object is destroyed (instead at `with` end). Also calling *close* twice (as it's public) would no longer crash.

As a side note, many tests failed when I ran them (original code).

**Recap** - this patch fixes:
- The double *close* crash
- Handle close performed at object destruction instead of the context manager's
- Better organize the code
- The 3<sup>rd</sup> commit excludes inherited handles from being closed (as it's each entity's responsibility to release resources that it allocated).

### DIFF
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -273,7 +273,7 @@ class AGraph(object):
     #        self.add_edge(u,v)
 
     def __del__(self):
-        self.close()
+        self._close_handle()
 
     def get_name(self):
         name = gv.agnameof(self.handle)
@@ -987,6 +987,7 @@ class AGraph(object):
         directed = self.directed
         self._close_handle()
         self.handle = gv.agraphnew(name, strict, directed)
+        self._owns_handle = True
         self._update_handle_references()
 
     def close(self):
@@ -1246,6 +1247,7 @@ class AGraph(object):
             except ValueError:
                 raise DotError("Invalid Input")
             else:
+                self._owns_handle = True
                 self._update_handle_references()
         except IOError:
             print("IO error reading file")
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -124,6 +124,7 @@ class AGraph(object):
                  filename=None, data=None, string=None, handle=None,
                  name='', strict=True, directed=False, **attr):
         self.handle = None  # assign first in case the __init__ bombs
+        self._owns_handle = True
         # initialization can take no arguments (gives empty graph) or
         # a file name
         # a string of graphviz dot language
@@ -158,6 +159,7 @@ class AGraph(object):
         if handle is not None:
             # if handle was specified, reference it
             self.handle = handle
+            self._owns_handle = False
         elif filename is not None:
             # load new graph from file (creates self.handle)
             self.read(filename)
@@ -990,12 +992,18 @@ class AGraph(object):
     def close(self):
         self._close_handle()
 
+
     def _close_handle(self):
         # may be useful to clean up graphviz data
         # this should completely remove all of the existing graphviz data
-        if self.handle is not None:
-            gv.agclose(self.handle)
+        if self._owns_handle:
+            if self.handle is not None:
+                gv.agclose(self.handle)
+                self.handle = None
+            self._owns_handle = False
+        else:
             self.handle = None
+
 
     def copy(self):
         """Return a copy of the graph."""
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -224,7 +224,7 @@ class AGraph(object):
         return self
 
     def __exit__(self, ext_type, exc_value, traceback):
-        self.close()
+        pass
 
     if _PY2:
         def __unicode__(self):
@@ -251,7 +251,6 @@ class AGraph(object):
         # hash the string representation for id
         return hash(self.string())
 
-
     def __iter__(self):
         # provide "for n in G"
         return self.nodes_iter()
@@ -270,6 +269,9 @@ class AGraph(object):
     # not implemented, but could be...
     #    def __setitem__(self,u,v):
     #        self.add_edge(u,v)
+
+    def __del__(self):
+        self.close()
 
     def get_name(self):
         name = gv.agnameof(self.handle)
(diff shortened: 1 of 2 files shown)


jarrodmillman:
See #213

dschult:
To handle the case when the handle is passed into the initializer -- and therefore we should not be closing it even when this object is deleted -- I think it would be sufficient to create a flag like ```self.handle_created_by_self``` which would be set to ```False``` when the handle isn't created during initialization -- that is, when it is passed into __init__.

Then we only close it if we created it.  @christifati  am I missing anything else that might cause trouble here?

CristiFati:
@dschult: I had the patch long ago but I didn't include it (as I'm not a *Graphviz* expert). Added it now. Things seem better than they were (I didn't test to see what happens when a shared handle gets closed by one instance). Also, updated the description.

Improvement ideas:

- Remove *close* altogether. There's no reason why one would use a graph whose inner handle was closed. But (client) code that already calls *close* will no longer work
- I don't know if someone is looking at the failed tests partial results, but (for my run) the ones that matter took 4 times less than the ones that failed. I'd recommend to disable the latter ones til they are fixed (and thus provide meaningful results)

CristiFati:
All: I don't know the proper way of asking a question (and definitely I wouldn't open a new issue for this one): what *Graphviz* version (latest: *v2.44.1*) would you recommend to use when building *v1.6* (on *Win*)?

dschult:
Version 2.40 has bugs for multiedges. We have successfully used v2.38 and 2.42 for a while now. v2.44 should work great too -- but I personally don't have much experience with it.  Still, I would suggest 2.44 because it's the latest.  For the third digit get the latest available.

(Even numbers are stable releases, odd numbers are development versions so that's why I don't talk about 2.41 or 2.43)

denisesato:
I am still getting the same problem with pygraphviz 1.7. The file handles are not correctly released on a Windows machine. Could someone help me? Should I change the version?

dschult:
What version of Graphviz do you have working with pygraphviz 1.7?

denisesato:
Graphviz 2.47.1

Em qua., 12 de mai. de 2021 às 10:51, Dan Schult ***@***.***>
escreveu:

dschult:
I believe it should be working with that combination of pygraphviz and graphviz, though I guess that 2.47 is an unstable release (an odd second number) and I haven't tried it myself. Can you recreate the problem with a small example?  Or maybe first try your same code with Graphviz 2.46?

denisesato:
I just create o log of AGraph objects based on a string describing one example graph.
After reading 509 graphs the application crashed. I can change the point of crashing by defining different maximum of allowed opened files (by using wfile._setmaxstdio). I left this line commented on the code. 
Based on this I assumed the problem is that the AGraph creates a temporary file to import the graph but `does` not correctly release it.


`import pygraphviz as pgv
import win32file as wfile

if __name__ == '__main__':

    #wfile._setmaxstdio(20)
    gviz_sample = """digraph {
                            graph [bgcolor=transparent]
                            node [shape=box]
                            -4955230706231096671 [label="advise claimant on reimbursement (20)" fillcolor="#FFFFFF" fontsize=12 style=filled]
                            3293026726485711842 [label="assess claim (30)" fillcolor="#CDCDFF" fontsize=12 style=filled]
                            6877596873952174207 [label="check if sufficient information is available (40)" fillcolor="#9B9BFF" fontsize=12 style=filled]
                            -5047037169028982727 [label="close claim (20)" fillcolor="#FFFFFF" fontsize=12 style=filled]
                            -2179017779376986680 [label="determine likelihood of claim (34)" fillcolor="#B9B9FF" fontsize=12 style=filled]
                            8939422555269194779 [label="end (40)" fillcolor="#9B9BFF" fontsize=12 style=filled]
                            5279819820230706915 [label="initiate payment (20)" fillcolor="#FFFFFF" fontsize=12 style=filled]
                            -6656769854897648909 [label="register claim (34)" fillcolor="#B9B9FF" fontsize=12 style=filled]
                            -4955230706231096671 -> -4955230706231096671 [label=10 fontsize=12 penwidth=1.7111107160496024]
                            -4955230706231096671 -> -5047037169028982727 [label=3 fontsize=12 penwidth=1.0888888395062002]
                            -4955230706231096671 -> 5279819820230706915 [label=7 fontsize=12 penwidth=1.4444441975310014]
                            3293026726485711842 -> -4955230706231096671 [label=7 fontsize=12 penwidth=1.4444441975310014]
                            3293026726485711842 -> 3293026726485711842 [label=15 fontsize=12 penwidth=2.1555549135806036]
                            3293026726485711842 -> 8939422555269194779 [label=5 fontsize=12 penwidth=1.2666665185186008]
                            3293026726485711842 -> 5279819820230706915 [label=3 fontsize=12 penwidth=1.0888888395062002]
                            6877596873952174207 -> 6877596873952174207 [label=20 fontsize=12 penwidth=2.5999991111116048]
                            6877596873952174207 -> 8939422555269194779 [label=3 fontsize=12 penwidth=1.0888888395062002]
                            6877596873952174207 -> -6656769854897648909 [label=17 fontsize=12 penwidth=2.3333325925930044]
                            -5047037169028982727 -> -5047037169028982727 [label=10 fontsize=12 penwidth=1.7111107160496024]
                            -5047037169028982727 -> 8939422555269194779 [label=10 fontsize=12 penwidth=1.7111107160496024]
                            -2179017779376986680 -> 3293026726485711842 [label=15 fontsize=12 penwidth=2.1555549135806036]
                            -2179017779376986680 -> -2179017779376986680 [label=17 fontsize=12 penwidth=2.3333325925930044]
                            -2179017779376986680 -> 8939422555269194779 [label=2 fontsize=12 penwidth=1.0]
                            8939422555269194779 -> 8939422555269194779 [label=20 fontsize=12 penwidth=2.5999991111116048]
                            5279819820230706915 -> -4955230706231096671 [label=3 fontsize=12 penwidth=1.0888888395062002]
                            5279819820230706915 -> -5047037169028982727 [label=7 fontsize=12 penwidth=1.4444441975310014]
                            5279819820230706915 -> 5279819820230706915 [label=10 fontsize=12 penwidth=1.7111107160496024]
                            -6656769854897648909 -> -2179017779376986680 [label=17 fontsize=12 penwidth=2.3333325925930044]
                            -6656769854897648909 -> -6656769854897648909 [label=17 fontsize=12 penwidth=2.3333325925930044]
                            overlap=false
                        }
                        """
    for i in range(0, 512):
        print(f'Reading {i} graph')
        graph = pgv.AGraph(gviz_sample)`

PR #234, opened 2020-08-02
OSX tests

dschult:
It seems that different combinations of operating systems and perhaps versions of AGraph lead to spacing issues which have always been a headache and don't help make the code better. 

So, I removed such issues with the " ".join(x.split()) metaphor.
Also, some systems put "" when the graph has no name while others
leave no text at all. 

I made a function stringify() to help process these issues.


dschult:
I think this makes osx testing work for pygraphviz.
But it no longer tests the precise spacing of AGraph.string().
I claim that changes from version to version and OS to OS so much
we shouldn't worry about it. It makes a valid dot file in any case.

If we are willing to give up that kind of test, I believe this is ready to go.
I suppose someday I should centralize ```stringify()``` to a testing or utils area. 
But I'd like to wait until more than a single function can be centralized. :}

PR #239, opened 2020-08-06
Remove stale mock dependency

mgorny:
Since pygraphviz no longer supports Python 2 and entirely relies
on built-in unittest.mock module, the dependency on mock is no longer
necessary.


dschult:
Thanks for this!

PR #240, opened 2020-08-06
Update doctests for Python 3, remove doctest-ignore-unicode

mgorny:
Since pygraphviz no longer supports Python 2, the unicode strings
in doctests are no longer meaningful.  Update the tests to use plain
Python 3 str, and remove doctest-ignore-unicode hack.

// This conflicts a bit with my other PR, I can rebase if that one is merged.

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -104,7 +104,7 @@ For Red Hat Enterprise Linux 7, the requirements are::
    
    sudo yum groups install -y "Development Tools"
    sudo yum install -y python-devel graphviz-devel
-   sudo python -m pip install nose doctest-ignore-unicode
+   sudo python -m pip install nose
 
 
 GraphViz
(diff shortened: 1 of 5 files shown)


dschult:
Ok... this looks good. Please rebase and we’ll clean this up. ;)

jarrodmillman:
I think we can also remove ``mock``.  I am not sure why it is being installed as I can't find any use of it.

jarrodmillman:
@mgorny Thanks for the two PRs!  I completely understand if you are busy with other things, but it would be great if you wanted to take on the move from nose to pytest (see #241).  Dan and I will be more than happy to provide feedback and answer questions.  Again, no worries if you are busy with other things.

mgorny:
Rebased now. Sure, I'll look at pytest. It shouldn't be hard.

dschult:
I think for NetworkX when we switched to pytest, we used ```pytest.importorskip``` for optional dependencies. Not sure there are many of those for pygraphviz. We also used ```pytest.raises()``` as well as ```with pytest.raises(Exception):``` a fair amount. Otherwise it should be straightforward (I think)... :}

Thanks!

PR #242, opened 2020-08-06
Use pytest for tests

mgorny:


.travis.yml
```diff
@@ -51,9 +51,9 @@ install:
   - pip list
 script:
   - if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
```
mgorny, line 53: TODO: check if this is still needed with pytest.

pygraphviz/__init__.py
```diff
@@ -56,6 +56,3 @@
 from .agraph import AGraph, Node, Edge, Attribute, ItemAttribute, DotError
 
 __all__ = ["AGraph", "Node", "Edge", "Attribute", "ItemAttribute", "DotError"]
-
-
-from pygraphviz.tests.test import run as test
```
mgorny, line 61: Do we still need this? For now I've removed it because it was imported everywhere.

setup.py
```diff
@@ -85,7 +85,6 @@
             },
         package_data=package_data,
         include_package_data = True,
-        test_suite='nose.collector',
```
mgorny, line 88: We could technically use pytest-runner here *but* setuptools upstream considers running `setup.py test` deprecated.

pygraphviz/__init__.py
dschult: I would suggest that we remove it. It allowed easy syntax for testing from an interpreter after import. But that's not the common way to test and it wasn't well known or advertised (I believe).

setup.py
dschult: Similar to the easy interface for testing after import, I don't think we need to provide ```setup.py test``` any longer.

mgorny:
Ok, I think it's ready for review. Travis on Linux passed, OSX is still waiting.

.travis.yml
dschult: We could try to do coverage for both osx and for linux, but I don't think it would be any different. Coverage shouldn't change much from system to system.  Either way I don't think nose was causing this choice, so pytest shouldn't make us change. I feel we can leave coverage turned off for osx to simplify and also to reduce an already (7x) longer execution time on the osx platform.

dschult:
Thanks very much!   This was so fast. :)   made my day! :}

PR #269, opened 2020-12-21
Use same theme as nx (for docs)

jarrodmillman:


rossbar (approved):
+1 for the new theme, the docs look very nice now: https://11-11846030-gh.circle-artifacts.com/0/doc/build/html/index.html

One idea for improving the landing page: add some introductory text to `index.rst` so that the whole page isn't only the toctree. Along with this, reducing the toctree either by setting `:maxdepth: 1` or even `:hidden:` so that the toctree isn't rendered on the landing page might be good (since the toctree is premanently embedded on the left sidebar with the rtd theme). This can be done in a followup though.

PR #353, opened 2021-05-30
docstring typos (paris -> pairs)

luketudge:
Saw this in the docs a couple of times. So here is a PR in case it is of use.


dschult:
Thanks for this!!   It looks like we were taking all the pairs to Paris!  :}

PR #382, opened 2022-01-10
make github actions happy, make py3.10 a string

MridulS:


MridulS:
Self merging, just a little change.

PR #394, opened 2022-02-01
Pass proper python version dependency in setup.py

vinayinvicible:


rossbar (approved):
Seems find to me unless we want to maintain "unofficial" support (i.e. not built or tested) for 3.7. Good find @vinayinvicible

jarrodmillman (approved):

jarrodmillman:
Thanks!  It looks like I missed this in #375.

PR #414, opened 2022-06-05
AGraph: Don't leak file descriptor on read("/path/")

nehaljwani:


jarrodmillman (approved):

rossbar (approved):
Thanks @nehaljwani ! I can't think of a straightforward way to test this and I don't think that should be a blocker. If anyone wants to follow up with a test case for this it'd be a nice improvement.

PR #430, opened 2022-08-09
Announce support for Python 3.11

jarrodmillman:
The 3.11.0 candidate 1 was released Monday, 2022-08-08. We've been testing against 3.11-dev for sometime and haven't seen any issues.

3.11.0 final is expected to be released on Monday, 2022-10-03. If we run into any 3.11 issues, we can address them and make a 1.10.1 release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,7 +88,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


jarrodmillman:
Waiting on https://github.com/actions/python-versions/pull/183.

PR #455, opened 2023-04-24
Update install instructions for usage with brew

jeanclaude-jardim-oxb:
Since pip 23.1 the`--global-option` and `--build-option` flags are being deprecated in favour of `config-settings`, more details at https://pip.pypa.io/en/stable/news/#v23-1.

This change aims to update the install instructions to work with pip versions from 23.1 and graphviz install via brew

Should resolve issue #454


rossbar (approved):
Thanks @jeanclaude-jardim-oxb !

PR #460, opened 2023-05-15
Add pep517 flag for config-settings

superpung:
Following a pip version update, the `--config-settings` feature will no longer be supported by default. As a result, an error message will be displayed suggesting alternative usage:

> DEPRECATION: Config settings are ignored for project pygraphviz from https://files.pythonhosted.org/packages/ee/7e/7366c082f959db7ee18a16fc38dc594158ede65ca789bef87751ed5130c7/pygraphviz-1.10.zip. pip 23.3 will enforce this behaviour change. A possible replacement is to use --use-pep517 or add a pyproject.toml file to the project.

In response, this pr adds an option `--use-pep517` that is documented within the installation instructions.

<details><summary>Screenshots</summary>
<h6>Before adding <code>--use-pep517</code></h6>
Before add --use-pep517
<h6>After adding <code>--use-pep517</code></h6>
After add --use-pep517
</details>

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -146,7 +146,7 @@ Manual download
 
 .. code-block:: console
 
-    PS C:\> python -m pip install --use-pep517
+    PS C:\> python -m pip install --use-pep517 `
                   --config-setting="--global-option=build_ext" `
                   --config-setting="--global-option="-IC:\Program Files\Graphviz\include" `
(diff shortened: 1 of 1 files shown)


jarrodmillman:
I am not seeing this deprecation warning with `pip == 23.1.2`, but it looks like the warning was added here:
- https://github.com/pypa/pip/pull/11919

More information here:
- https://github.com/pypa/pip/issues/11915

PR #465, opened 2023-06-01
Drop Python 3.8 support per SPEC 0

jarrodmillman:


rossbar (approved):
Thanks @jarrodmillman

PR #473, opened 2023-08-10
Copy edges keys when copying a graph

### LINKED ISSUE #472: Edge keys not preserved when copying multigraphs using copy() method
`AGraph` copying logic for multigraphs, using `copy()`, skips edges keys.

```Python
import pygraphviz as pgv

# Create a graph and add an edge with an explicit key
A = pgv.AGraph(directed=True, strict=False)
A.add_edge(1, 2, key='edge1')

# Copy the graph
B = A.copy()

# Print edge keys for both graphs
print("Original graph edge keys:", [k for s, t, k in A.edges(keys=True)])
print("Copied graph edge keys:", [k for s, t, k in B.edges(keys=True)])
```

This results with:
```
Original graph edge keys: ['edge1']
Copied graph edge keys: [None]
```

The expected behavior is that the keys in the copied graph will be identical to the original graph.

elishapiiro:
Fix #472 

Some attributes are not copied when copying a graph, like edges keys. This fix copies edge keys to the target graph so it matches the original graph.

### DIFF
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -1027,7 +1027,7 @@ class AGraph:
         for node in self.nodes():
             G.add_node(node)
             G.get_node(node).attr.update(self.get_node(node).attr)
-        for edge in self.edges():
+        for edge in self.edges(keys=True):
             G.add_edge(*edge)
             G.get_edge(*edge).attr.update(self.get_edge(*edge).attr)
         G.graph_attr.update(self.graph_attr)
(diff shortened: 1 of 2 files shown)


dschult (approved):
Thank you for this!   If you find other places where multigraphs are undersupported we'd appreciate updates.

This looks good to me.

rossbar (approved):
Very nice fix + clear test, thanks @elishapiiro !

PR #484, opened 2023-11-22
Update INSTALL.txt

moemen95:
Fix Typo in pip command to "--config-settings" Reference: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-C.

This command fixed my issue on installing pygraphviz on my mac M1 Chip. But we need to fix the typo of that option in the docs, to help other developers.


michaelbeale-IL:
I stumbled across this PR trying to get pygraphviz installed on my Mac.  The change worked for me, but only after I explicitly installed pep517 in my virtual environment.  Not sure if that helps.

PR #508, opened 2024-01-25
Remove outdated pystrings.swg

wsfulton:
Remove pystring.swg - a very ancient copy from SWIG which does not compile when using swig-4.2.0. Instead SWIG's version of pystrings.swg that is kept up to date in SWIG is used. SWIG_PYTHON_STRICT_BYTE_CHAR needs to be defined in order to maintain the current behaviour of only allowing Python 3 byte type instead of Python 3 string type as input.

As the output of swig is in graphviz_wrap.c and is committed to the repo, the updates committed in this file (keeping the current swig-4.1.1 version). A later commit could update it to swig-4.2.0.

Fixes build problems on Fedora where SWIG_Python_str_AsChar no longer exists, as reported at https://github.com/swig/swig/issues/2778.


jarrodmillman (approved):

jarrodmillman:
Thanks @wsfulton !!

PR #551, opened 2025-02-03
DOC: remove pip install as recommendation

MridulS:
We shouldn't put pip as an explicit recommendation if we need people to use custom pip config for Windows and MacOS. This doesn't recommend conda either, just removes the explicit recommendation.


rossbar (approved):
Agreed, thanks @MridulS !

PR #559, opened 2025-02-12
More precise allocation for string copy in SWIG glue

Smattr:
A few minor commits, then the focus of this series in the final one whose commit message I quote below.

----

This code was over-allocating and then adjusting string content to truncate the copied string. This has a couple of minor weaknesses:
  1. One more byte than necessary is allocated, increasing heap pressure.
  2. Memory safety tools like Address Sanitizer cannot spot 1-byte overruns.

(1) is not significant, but (2) can make it harder to debug memory safety problems. This change tightens the bounds:

```
                           ┌───┬───┬───┬───┬───┬───┐
  original val:            │ < │ f │ o │ o │ > │000│
                           └───┴───┴───┴───┴───┴───┘
                               ┌───┬───┬───┬───┬───┐
  copy before this change:     │ f │ o │ o │000│000│
                               └───┴───┴───┴───┴───┘
                               ┌───┬───┬───┬───┐ ▲
  copy after this change:      │ f │ o │ o │000│ │
                               └───┴───┴───┴───┘ │
                                                 │
                                ASan, Valgrind etc previously did not know
                                  reading or writing this byte should be
                                    considered an out-of-bounds access
```

This change has no effect on unchecked allocation failures in this code. In future, this should probably be improved to check for and handle `NULL` being returned from `malloc`.

### DIFF
--- a/pygraphviz/graphviz.i
+++ b/pygraphviz/graphviz.i
@@ -7,6 +7,7 @@
 %{
 #include "graphviz/cgraph.h"
 #include "graphviz/gvc.h"
+#include <stddef.h>
 %}
 
 %typemap(in) FILE* input_file (int fd, PyObject *mode_obj, PyObject *mode_byte_obj, char *mode) {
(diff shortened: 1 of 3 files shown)


rossbar (approved):
Thanks @Smattr for the changes and the explanation - LGTM. I don't think there's anything in the test suite that would probe this but everything looks good. CI failures unrelated!

jarrodmillman (approved):
LGTM

PR #566, opened 2025-06-16
Fix types

schaeferpp:
On my machine, running arch linux with graphviz version 13.0.0-1, I got the following error on building pygraphviz.

```
pygraphviz/graphviz_wrap.c: In function ‘_wrap_gvRenderData’:
      pygraphviz/graphviz_wrap.c:5480:50: error: passing argument 5 of ‘gvRenderData’ from incompatible pointer type [-Wincompatible-pointer-types]
       5480 |   result = (int)gvRenderData(arg1,arg2,arg3,arg4,arg5);
            |                                                  ^~~~
            |                                                  |
            |                                                  unsigned int *
      In file included from pygraphviz/graphviz_wrap.c:3024:
      /usr/include/graphviz/gvc.h:99:49: note: expected ‘size_t *’ {aka ‘long unsigned int *’} but argument is of type ‘unsigned int *’
         99 |                          char **result, size_t *length);
            |                                         ~~~~~~~~^~~~~~
      error: command '/usr/bin/gcc' failed with exit code 1
```

Most likely it would be better to just regenerate the file for graphviz 13, but this just works.

### DIFF
--- a/pygraphviz/graphviz_wrap.c
+++ b/pygraphviz/graphviz_wrap.c
@@ -5447,7 +5447,7 @@ SWIGINTERN PyObject *_wrap_gvRenderData(PyObject *self, PyObject *args) {
   Agraph_t *arg2 = (Agraph_t *) 0 ;
   char *arg3 = (char *) 0 ;
   char **arg4 = (char **) 0 ;
-  unsigned int *arg5 = (unsigned int *) 0 ;
+  size_t *arg5 = (size_t *) 0 ;
   void *argp1 = 0 ;
   int res1 = 0 ;
   void *argp2 = 0 ;
@@ -5456,7 +5456,7 @@ SWIGINTERN PyObject *_wrap_gvRenderData(PyObject *self, PyObject *args) {
   char *buf3 = 0 ;
   int alloc3 = 0 ;
   char *temp4 = 0 ;
-  unsigned int tempn4 ;
+  size_t tempn4 ;
   PyObject *swig_obj[3] ;
   int result;
   


rossbar (commented):
It looks like there are underlying changes that are not compatible between graphviz v12 and v13 and this will break things for v12, right? If so, we're going to have to find a way to support both.

kylefhartzenberg:
I am experiencing the same issue when attempting to `pip install pygraphviz` on Arch Linux with Python 3.15.5 and gcc 15.1.1 20250425 inside a virtual environment (see output below). Is this patch going to be applied to the Python Package Index (PyPI)  release anytime soon?

I have successfully installed and used PyGraphviz provided by the Arch Linux package repository where [this patch](https://gitlab.archlinux.org/archlinux/packaging/packages/python-pygraphviz/-/commit/1c94b3a0c6be99900e224d43c98f28cf66825a00) has been applied.

**Output**
```none
creating build/temp.linux-x86_64-cpython-313/pygraphviz
gcc -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O3 -Wall -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -ffat-lto-objects -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -fPIC -DSWIG_PYTHON_STRICT_BYTE_CHAR -I/home/user/dev/py/env/lab/include -I/usr/include/python3.13 -c pygraphviz/graphviz_wrap.c -o build/temp.linux-x86_64-cpython-313/pygraphviz/graphviz_wrap.o
pygraphviz/graphviz_wrap.c:9:9: warning: ‘SWIG_PYTHON_STRICT_BYTE_CHAR’ redefined
    9 | #define SWIG_PYTHON_STRICT_BYTE_CHAR
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
<command-line>: note: this is the location of the previous definition
pygraphviz/graphviz_wrap.c: In function ‘_wrap_gvRenderData’:
pygraphviz/graphviz_wrap.c:5480:50: error: passing argument 5 of ‘gvRenderData’ from incompatible pointer type [-Wincompatible-pointer-types]
 5480 |   result = (int)gvRenderData(arg1,arg2,arg3,arg4,arg5);
      |                                                  ^~~~
      |                                                  |
      |                                                  unsigned int *
In file included from pygraphviz/graphviz_wrap.c:3024:
 /usr/include/graphviz/gvc.h:99:49: note: expected ‘size_t *’ {aka ‘long unsigned int *’} but argument is of type ‘unsigned int *’
   99 |                          char **result, size_t *length);
      |                                         ~~~~~~~~^~~~~~
error: command '/usr/bin/gcc' failed with exit code 1
```

opoplawski:
This patch also is only changing the generated .c file, but not the original .i source.  I am also seeing test failures:
```
__ TestExperimentalGraphvizLibInterface.test_drawing_no_error_with_no_layout ___

self = <pygraphviz.tests.test_drawing.TestExperimentalGraphvizLibInterface object at 0x7f987f402d70>

    def test_drawing_no_error_with_no_layout(self):
        A.add_path([1, 2, 3, 4])
>       d = A._draw(prog="nop")

../BUILDROOT/usr/lib64/python3.14/site-packages/pygraphviz/tests/test_drawing.py:104:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <AGraph b'test graph' <Swig Object of type 'Agraph_t *' at 0x7f987f29c510>>
path = None, format = b'dot', prog = b'nop', args = ''

    def _draw(self, path=None, format=None, prog=None, args=""):
        """Output graph to path in specified format.

        .. caution:: EXPERIMENTAL

        This version of the draw command uses libgvc for drawing instead
        of command line GraphViz tools like in versions <1.6 and the default.

        An attempt will be made to guess the output format based on the file
        extension of `path`.  If that fails, then the `format` parameter will
        be used.

        Note, if `path` is a file object returned by a call to os.fdopen(),
        then the method for discovering the format will not work.  In such
        cases, one should explicitly set the `format` parameter; otherwise, it
        will default to 'dot'.

        If path is None, the result is returned as a Bytes object.

        Formats (not all may be available on every system depending on
        how Graphviz was built)

            'canon', 'cmap', 'cmapx', 'cmapx_np', 'dia', 'dot',
            'fig', 'gd', 'gd2', 'gif', 'hpgl', 'imap', 'imap_np',
            'ismap', 'jpe', 'jpeg', 'jpg', 'mif', 'mp', 'pcl', 'pdf',
            'pic', 'plain', 'plain-ext', 'png', 'ps', 'ps2', 'svg',
            'svgz', 'vml', 'vmlz', 'vrml', 'vtx', 'wbmp', 'xdot', 'xlib'


        If prog is not specified and the graph has positions
        (see layout()) then no additional graph positioning will
        be performed.

        Optional prog=['neato'|'dot'|'twopi'|'circo'|'fdp'|'nop']
        will use specified graphviz layout method.

        >>> import pygraphviz as pgv
        >>> G = pgv.AGraph()
        >>> G.add_edges_from([(0, 1), (1, 2), (2, 0), (2, 3)])
        >>> G.layout()

        # use current node positions, output pdf in 'file.pdf'
        >>> G.draw("file.pdf")

        # use dot to position, output png in 'file'
        >>> G.draw("file", format="png", prog="dot")

        # use keyword 'args' to pass additional arguments to graphviz
        >>> G.draw("test.pdf", prog="twopi", args="-Gepsilon=1")
        >>> G.draw("test2.pdf", args="-Nshape=box -Edir=forward -Ecolor=red ")

        The layout might take a long time on large graphs.

        """
        # try to guess format from extension
        if format is None and path is not None:
            p = path
            # in case we got a file handle get its name instead
            if not isinstance(p, str):
                p = path.name
            format = os.path.splitext(p)[-1].lower()[1:]

        if format is None or format == "":
            format = "dot"

        if prog is None:
            if self.has_layout:
                prog = "neato"
                args += " -n2"
            else:
                msg = """Graph has no layout information, see layout() or specify prog={}.""".format(
                    "|".join(["neato", "dot", "twopi", "circo", "fdp", "nop"])
                )
                raise AttributeError(msg)

        else:
            if self.number_of_nodes() > 1000:
                sys.stderr.write(
                    f"Warning: graph has {self.number_of_nodes()} nodes...layout may take a long time.\n"
                )

        # process args
        format, prog = self._manually_parse_args(args, format, prog)

        # convert input strings to type bytes (encode it)
        if isinstance(format, str):
            format = format.encode(self.encoding)
        if isinstance(prog, str):
            prog = prog.encode(self.encoding)

        # Start the drawing
        gvc = gv.gvContext()
        G = self.handle

        # Layout
        err = gv.gvLayout(gvc, G, prog)
        if err:
            if err != -1:
                raise ValueError("Graphviz raised a layout error.")
            prog = prog.decode(self.encoding)
            raise ValueError(f"Can't find prog={prog} in this graphviz installation")

        # Render
        if path is None:
>           out = gv.gvRenderData(gvc, G, format)
E           TypeError: gvRenderData() missing 2 required positional arguments: 'result' and 'size'

../BUILDROOT/usr/lib64/python3.14/site-packages/pygraphviz/agraph.py:1724: TypeError
```

muety:
Any chances to get this merged soon?

Smattr:
At a guess, I would say #573 might be the “proper” way to fix this.

(I believe the maintainers are aware of this, but for those following along at home who want the backstory… [Graphviz commit cd3beefc4555f505445582268de5e3c53e37b0e3](https://gitlab.com/graphviz/graphviz/-/commit/cd3beefc4555f505445582268de5e3c53e37b0e3) changed the type of `gvRenderData`. This landed in [Graphviz 13.0.0](https://gitlab.com/graphviz/graphviz/-/blob/b267806f1e18f2b00fd88b9c3157461df4d2e07f/CHANGELOG.md#1300--2025-06-08) as an API break. It seems pygraphviz does not have an established mechanism to handle Graphviz API breaks, which served it well because the Graphviz API went a long time with no meaningful breaks. I am not sure exactly where/how it makes most sense to introduce some API compatibility management in pygraphviz.)

rossbar:
Thanks @Smattr for the context and perspective! Yes pygraphviz folks are indeed aware of the break but haven't had a bandwidth to get a general fix for all graphviz versions out (yet).

>  I am not sure exactly where/how it makes most sense to introduce some API compatibility management in pygraphviz.

In principle, the ideal solution is to wrap the subset of graphviz libraries that pygraphviz depends on (e.g. libcgraph, libcdt, etc.) up into the wheel directly, so we can package whichever version of graphviz we need along with pygraphviz and can do away with adding scaffolding for detecting graphviz major versions. We've had some recent contributions that look promising in this direction - we'll see if we can make some progress!

rossbar:
With #583 merged, this problem will (once a release is cut and the wheels are published) be solved for users; i.e. folks who are `pip install` ing the wheels. The great thing about getting wheels out is it incorporates all of the graphviz dependencies inside the wheel itself, so there is no longer any build/install-time dependency on graphviz on users' machines. This has been an *extremely* longstanding goal.

However - this does *not* solve the issue for packagers - e.g. folks who maintain the `pygraphviz` packages for linux distros. Those packages are independent of wheels and will still have the issue of which version of graphviz they support/depend on. Let's continue that discussion in #573

rossbar:
This was fixed in #603 following @mcepl 's proposal to keep support for both old (pre-v13) and new (post-v13) graphviz versions.

PR #573, opened 2025-10-22
fix: correct signature of gvRenderData() function

### LINKED ISSUE #567: Type Error:  expected ‘size_t *’ {aka ‘long unsigned int *’} but argument is of type ‘unsigned int *’
The version of graphviz is `13.0.1`.

The error message is as below:

```
 pygraphviz/graphviz_wrap.c: In function ‘_wrap_gvRenderData’:
      pygraphviz/graphviz_wrap.c:5480:50: error: passing argument 5 of ‘gvRenderData’ from incompatible pointer type [-Wincompatible-pointer-types]
       5480 |   result = (int)gvRenderData(arg1,arg2,arg3,arg4,arg5);
            |                                                  ^~~~
            |                                                  |
            |                                                  unsigned int *
      In file included from pygraphviz/graphviz_wrap.c:3024:
      /usr/include/graphviz/gvc.h:99:49: note: expected ‘size_t *’ {aka ‘long unsigned int *’} but argument is of type ‘unsigned int *’
         99 |                          char **result, size_t *length);
            |                                         ~~~~~~~~^~~~~~
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]

```

`arg5` here should be with type `size_t *` instead of `unsigned int *`.

mcepl:
For compatibility with graphviz 14.*

Yes, it adds `swig` as build time dependency, because I think it is The Right Thing™ to do to use the original source. However, it can be changed into using generated C file (which should be then properly packaged).

Fixes: https://github.com/pygraphviz/pygraphviz/issues/567
Replaces: https://github.com/pygraphviz/pygraphviz/pull/566


Smattr:
This feels like the right direction to me. Though unfortunately the break occurred in 13.0.0 and the `GRAPHVIZ_MAJOR_VERSION` macro was only introduced in 14.0.0. If you want more robust behaviour across [13.0.0, 14.0.0) and can rely on the binaries being installed, you could try something like:

```python
match = re.match(
  r"dot - graphviz version (\d+)",
  subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT)
)
```

rossbar:
Okay - as mentioned in #567 , the graphviz dependency problem is fixed (or will be after release) for folks installing from wheels.

For folks installing from source however, we still have this problem. Having thought about this a bit, 2 approaches occur to me:

1. Dynamic version-detection at build-time; i.e. something akin to what's proposed here.
2. A lower-bound pin on graphviz versions

Which one makes the most sense depends on the packagers' needs. IMO option 2 is preferable for pygraphviz - it obviates the need for version-based switches in the code and adding infrastructure to test against different graphviz versions. However, if there is a strong need to support many different versions of graphviz, then that changes the picture. Thoughts?

rossbar:
Thanks again @mcepl and @Smattr - the core here were incorporated in #603 (with attribution) so I'll go ahead and close this one!

PR #582, opened 2026-02-10
Drop support for python 3.11 (SPEC 0)

MridulS:


rossbar (approved):
Thanks @MridulS 

In principle we could backport support for earlier versions of Python (say to 3.10) for the very first release of Pygraphviz 2.0 once the windows wheel problem is finally solved, but that's a special case - +1 for SPEC 0 for development!

PR #555, opened 2025-02-07
DRAFT: cibuildwheel take 3

MridulS:

PR #553, opened 2025-02-05
DRAFT: cibuildwheel

MridulS:
This forks off https://github.com/pygraphviz/pygraphviz/pull/537 just to rerun the runners and look at the current state with newer versions of graphviz.

PR #583, opened 2026-02-10
CI: cibuildwheel take 4

MridulS:
This builds on top of https://github.com/pygraphviz/pygraphviz/pull/553, https://github.com/pygraphviz/pygraphviz/pull/555, https://github.com/pygraphviz/pygraphviz/pull/537 and https://github.com/pygraphviz/pygraphviz/pull/558 with lots of LLM poking.


rossbar (commented):
Thanks @MridulS !

From a quick review, the high-level take is that this differs a bit from the approach in #558 --- rather than try to use graphviz's library interface and package the necessary libs inside the wheel (which works for all platforms except windows, due to the [library symbol export issue ID'ed there](https://github.com/pygraphviz/pygraphviz/pull/558#issuecomment-2644715286)); this approach instead just packages the necessary CLI scripts in the wheel then [does some path munging](https://github.com/pygraphviz/pygraphviz/pull/583/changes#diff-8a35a388dff56c1371ff8470ae6fca0f6bc46fa0cc5530f1ad0c7eb016ee27d6R1696) to make sure those CLI tools are discoverable at runtime.

I'm not crazy about the mess-with-users'-path-under-the-hood pattern, but it *does* avoid the library issue and allow for a fully self-contained wheel! Since this is the absolute number 1 issue with pygraphviz (and the root of basically every open issue), I think any approach that solves it is worthwhile!

That said, I'm also curious to see if some of the library discovery/build patches will help in #558 - e.g. [this chunk](https://github.com/pygraphviz/pygraphviz/pull/583/changes#diff-315dca192cfac8853e3fe5c956c6988a105d7ea3a60095eb00a1aa9d6516eb76R352) looks promising!

One final thought: the CIBW recipes for linux/macos are overprescribed I think - from experimenting in #558 , I recall that installing graphviz from the default pkg manager (along with the -devel pkg) in the `manylinux_2_34` image is just fine for wrapping. Some of the non-dll-export flags in these build recipes seem counterproductive.

Overall - I'm very +1 for pursuing this avenue!

MridulS:
> I'm not crazy about the mess-with-users'-path-under-the-hood pattern, but it does avoid the library issue and allow for a fully self-contained wheel! Since this is the absolute number 1 issue with pygraphviz (and the root of basically every open issue), I think any approach that solves it is worthwhile!

Yes, I'm leaning towards getting this self contained wheel as it also gets rid of the graphviz version dances I think. My strategy here was to turn off tests and then keep on turning them on with all the plugins. So I guess that's how I ended with this.

> One final thought: the CIBW recipes for linux/macos are overprescribed I think - from experimenting in https://github.com/pygraphviz/pygraphviz/pull/558 , I recall that installing graphviz from the default pkg manager (along with the -devel pkg) in the manylinux_2_34 image is just fine for wrapping. Some of the non-dll-export flags in these build recipes seem counterproductive.

Yeah, I need to look at this again. 

I lost my comment somewhere in the github cache but another issue here is the macosx_deployment_target is 15.0 which is too high IMO. The current setup is using the png libs from homebrew which have been compiled for 15.0 and we would need to recompile them in the current setup, so need to a better answer here.

dschult:
This looks good -- Do the things to look at (mentioned in comments) need to be looked at before this PR is merged? O are they OK for a followup?

Thanks for this!

rossbar (approved):
Looking at this again I think we should get it in!

The issue with packaging the CLI tools in the wheel is actually more muted than I originally thought as it's only for the less-commonly-used features (e.g. `tred`). The primary layouts (i.e. what's used in the NetworkX gallery) are indeed picked up correctly by the library interface! In principle, that means the graphviz API breakage issue *shouldn't* affect users who are using the more common layout functions (`neato`, `dot`, `circo` etc.).

Given the current state of things (API breakages in graphviz, broken packages on conda-forge, etc.) I'm in favor of pushing this forward! My previous concerns (e.g. overprescribed build recipes) pale in comparison to the fact that we're broken in users space now :)

Thanks again @MridulS for putting this together and @robtaylor for the helpful prods along these lines as well - that provided useful context for me specifically for the windows-specific bits.

dschult (approved):
This looks good to me!

PR #587, opened 2026-04-03
Use gvContextPlugins with platform-specific plugin loading

robtaylor:
## Summary

Builds on the approach from #558 to switch from `gvContext()` to `gvContextPlugins()` with builtin plugin symbols, enabling self-contained layout/render (prerequisite for wheels, #545).

The key change: platform-specific plugin loading that fixes the Windows `LNK2001` linker errors from #558.

### How it works

- **Unix (Linux/macOS)**: Plugin symbols (`gvplugin_core`, `gvplugin_dot_layout`, `gvplugin_neato_layout`) are resolved at link time via `extern` declarations and passed to `gvContextPlugins()` in a static `lt_preloaded_symbols` table.
- **Windows**: Plugin DLLs don't export `_LTX_library` symbols through import libraries, so link-time resolution fails. Instead, `LoadLibrary`/`GetProcAddress` resolves the symbols at runtime. Falls back to `gvContext()` if the DLLs aren't accessible.

### Implementation details

- `pygraphviz_context()` C helper function encapsulates the platform logic — no need to expose `lt_preloaded_symbols` as a Python-level `cvar`
- `extern` declarations placed in a plain `%{...%}` SWIG block (not `%inline`) so SWIG doesn't generate Python-level accessors that reference the symbols unconditionally
- Plugin libraries only added to the linker on non-Windows in `setup.py`
- Regenerated `graphviz_wrap.c` with SWIG 4.4.0

### CI fixes (separate commits)

- Added `$ErrorActionPreference = 'Stop'` to Windows install step so `pip install` failures properly fail the step
- Limited `push` trigger to `main` branch to avoid duplicate runs on PRs; added concurrency group

## Test plan

- [x] All 163 tests pass locally on macOS
- [x] CI passes: Ubuntu (3.12, 3.13, 3.14), Fedora (40, 41), brew (3.12, 3.13, 3.14), macports (3.12, 3.13, 3.14), **Windows (3.12, 3.13, 3.14)**

See passing CI on fork: https://github.com/robtaylor/pygraphviz/pull/1


robtaylor:
Closing in favor of #583 which takes a cleaner approach (`gvContext()` + `gvAddLibrary()` + `__declspec(dllimport)`) and includes the full cibuildwheel infrastructure.

Will open a separate small PR for the CI improvements (ErrorActionPreference fix and duplicate run prevention) that apply independently.

PR #591, opened 2026-04-22
Add aarch64 wheels for linux to cibuildwheel.

rossbar:
An attempt at adding aarch64 wheels for linux to the cibuildwheel action.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -25,6 +25,7 @@ jobs:
             macos-15-intel,
             windows-latest,
             ubuntu-24.04-arm,
+            windows-11-arm,
           ]
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


rossbar:
Note - just learned about [native arm runners](https://github.blog/changelog/2024-06-24-github-actions-ubuntu-24-04-image-now-available-for-arm64-runners/) so will switch to that - I just wanted to see if this would work (and get aarch wheels from the upload action to test locally).

rossbar:
Good news - wheels on linux-aarch work as expected! Converting to draft now as I don't have time to continue ATM. Will replace emulation with the native runners then this should be good to go!

MridulS (approved):

MridulS:
Added the linux arm machine in the matrix, should we also add windows-arm ?

rossbar:
> Added the linux arm machine in the matrix, should we also add windows-arm ?

Just gave this a shot in 6bdfd6d - it doesn't work out-of-the-box... it seems the blocker is building graphviz itself on arm windows. Let's look at that separately in a followup!

rossbar:
Thanks for getting the native runners in @MridulS - it's super convenient that cibuildwheel detects the architecture and just does the right thing :)

PR #592, opened 2026-04-30
Bump minimum supported graphviz version to v13

rossbar:
This is a "needs decision" for the pygraphviz 2.0 release.

This PR implements the alternative to dynamic build-time version checking proposed in #573.

With this approach, pygraphviz could only be built *from source* with graphviz>=13 installed.


rossbar:
Okay the build recipe for graphviz on macos doesn't work with the latest graphviz. Looking at the logs I suspect [this is the culprit](https://github.com/pygraphviz/pygraphviz/blob/ca8ad7c382f709bc2d1059461465a5910d43603d/.github/workflows/cibuildwheel.yml#L111).

rossbar:
We went with @mcepl 's more flexible proposal to continue supporting pre-v13 graphviz as well. Closing!

PR #603, opened 2026-06-06
Graphviz support across v13 API change

rossbar:
This PR combines @mcepl and @Smattr 's work/suggestions to handle the API change in graphviz 13. The intent here is to come up with the minimal change set that gets pygraphviz building again across all graphviz versions (>2.38).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,10 +14,15 @@ if __name__ == "__main__":
     # If, in the future, it is possible/reasonable to set Graphviz 14 as a
     # minimum supported version, then this should be replaced with the macro
     # See gh-573 for further discussion
-    vm = re.match(
-        r"dot - graphviz version (\d+)",
-        subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT).decode()
-    )
+    try:  # Better exception message if graphviz not installed
+        version_str = subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT)
+    except FileNotFoundError:
+        raise FileNotFoundError(
+            "\n\nCould not find `dot` executable!\n"
+            "Graphviz must be installed to build pygraphviz from source.\n\n"
+        )
+
+    vm = re.match(r"dot - graphviz version (\d+)", version_str.decode())
     graphviz_major_version = int(vm.string[:vm.end()].split(" ")[-1])
     print(f"Detected Graphviz version {graphviz_major_version}")
     # Pass version info into swig build
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ if __name__ == "__main__":
         )
 
     vm = re.match(r"dot - graphviz version (\d+)", version_str.decode())
-    graphviz_major_version = int(vm.string[:vm.end()].split(" ")[-1])
+    graphviz_major_version = int(vm.string[: vm.end()].split(" ")[-1])
     print(f"Detected Graphviz version {graphviz_major_version}")
     # Pass version info into swig build
     swig_options = [f"-DGRAPHVIZ_VERSION_MAJOR={graphviz_major_version}"]
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,12 @@ from setuptools import setup, Extension
 if __name__ == "__main__":
     WINDOWS = sys.platform == "win32"
 
+    # FIXME: This should be handled with the cibuildwheel windows path specification
+    # but for whatever reason it's not being picked up correctly. This is a hack
+    # to make this work with the specific installation locations used within the
+    # cibuildwheel environment.
+    if WINDOWS:
+        os.environ["PATH"] += ";C:\\graphviz\\bin"
     # Extract graphviz major version from command line
     # NOTE: A GRAPHVIZ_MAJOR_VERSION macro was added in Graphviz v14.0, but is
     # not availble prior - so parsing the CLI version output is the only
@@ -14,8 +20,6 @@ if __name__ == "__main__":
     # If, in the future, it is possible/reasonable to set Graphviz 14 as a
     # minimum supported version, then this should be replaced with the macro
     # See gh-573 for further discussion
-    if WINDOWS:
-        os.environ["PATH"] += ";C:\\graphviz\\bin"
     try:  # Better exception message if graphviz not installed
         version_str = subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT)
     except FileNotFoundError:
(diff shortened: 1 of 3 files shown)


Smattr:
Thanks for pushing this forwards.

Apologies for not thinking of this earlier, but there’s actually unrelated macro changes in Graphviz 13.0.0 that could be leaned on to detect this:

```
$ cat foo.c
#include <cgraph.h>

#ifdef agnew // code for Graphviz < 13
foo
#else // code for Graphviz ≥ 13
bar
#endif
$ gcc -E -I/tmp/graphviz-12.1.1/include/graphviz foo.c
…
# 2 "foo.c" 2


foo
$ gcc -E -I/tmp/graphviz-13.0.0/include/graphviz foo.c
…
# 2 "foo.c" 2




bar
```

Do you think using something like this would be simpler?

----

On a separate topic, what happens if someone installs Pygraphviz with a pre-existing <v13 installation of Graphviz, then later upgrades their Graphviz installation past v13? Will Pygraphviz then be silently calling into Graphviz with the wrong ABI? Is this a supported use case?

rossbar:
Thanks for the pointer @Smattr - that may indeed be simpler. My plan will be to try to take this PR to conclusion, then try a new PR to see if we can use this pattern instead. The main benefit would be getting rid of the dependency on `dot` at build time, but that likely only affects a small group of users (e.g. distro package maintainers)

> On a separate topic, what happens if someone installs Pygraphviz with a pre-existing <v13 installation of Graphviz, then later upgrades their Graphviz installation past v13? Will Pygraphviz then be silently calling into Graphviz with the wrong ABI? Is this a supported use case?

Good question - for the typical user (i.e. those who install `pygraphviz` from `pip`), the graphviz libraries will actually be packaged inside the wheel, so there is no external dependency on graphviz itself. Therefore the graphviz installation is completely decoupled from the pygraphviz installation and there should be no conflicts. Similarly, users who use their distro's package manager should be fine (as the distro packagers take on the task of dependency management, and upgrading graphviz should trigger an upgrade in the pygraphviz dependency).

I guess we'll see how noisy it is, but I suspect it's a corner case most folks won't hit (famous last words...)

MridulS (approved):

Smattr:
> My plan will be to try to take this PR to conclusion, then try a new PR to see if we can use this pattern instead.

OK, sounds good.

> Good question - for the typical user (i.e. those who install `pygraphviz` from `pip`), the graphviz libraries will actually be packaged inside the wheel, so there is no external dependency on graphviz itself. Therefore the graphviz installation is completely decoupled from the pygraphviz installation and there should be no conflicts. Similarly, users who use their distro's package manager should be fine (as the distro packagers take on the task of dependency management, and upgrading graphviz should trigger an upgrade in the pygraphviz dependency).

This surprises me because I thought the root cause of #566 was users BYOing an uncontrolled version of Graphviz. If the only supported workflow is Graphviz libraries being bundled with the Pygraphviz installation, just drop support for Graphviz < 13 and only ever bundle Graphviz ≥ 13. What am I missing?

setup.py
```diff
@@ -1,10 +1,39 @@
...
 
+    # FIXME: This should be handled with the cibuildwheel windows path specification
+    # but for whatever reason it's not being picked up correctly. This is a hack
+    # to make this work with the specific installation locations used within the
+    # cibuildwheel environment.
+    if WINDOWS:
+        os.environ["PATH"] += ";C:\\graphviz\\bin"
+    # Extract graphviz major version from command line
+    # NOTE: A GRAPHVIZ_MAJOR_VERSION macro was added in Graphviz v14.0, but is
+    # not availble prior - so parsing the CLI version output is the only
```
Smattr, line 18: nit: s/availble/available

setup.py
```diff
@@ -1,10 +1,39 @@
...
+    # If, in the future, it is possible/reasonable to set Graphviz 14 as a
+    # minimum supported version, then this should be replaced with the macro
+    # See gh-573 for further discussion
+    try:  # Better exception message if graphviz not installed
+        version_str = subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT)
+    except FileNotFoundError:
+        raise FileNotFoundError(
+            "\n\nCould not find `dot` executable!\n"
+            "Graphviz must be installed to build pygraphviz from source.\n\n"
+        )
```
Smattr, line 29: nit: you could do something like the following to also retain all info from the original exception:

```python
except FileNotFoundError as e:
  raise FileNotFoundError(…) from e
```

setup.py
```diff
@@ -1,10 +1,39 @@
...
+
+    vm = re.match(r"dot - graphviz version (\d+)", version_str.decode())
+    graphviz_major_version = int(vm.string[: vm.end()].split(" ")[-1])
```
Smattr, line 32: Instead of the above and this, I would do:

```python
try:
  version_str = subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT, text=True)
except FileNotFoundError:
  …
vm = re.match(r"dot - graphviz version (?P<major>\d+)", version_str)
graphviz_major_version = int(vm.group("major"))
```

or possible even omit the `int(` given nothing later requires on this variable being an `int`

Smattr (commented):
Not sure if post-merge review is helpful, but I left a couple of minor improvement suggestions.

setup.py
rossbar: True - I added the `int` as another implicit check to make sure I'm not accidentally picking up other parts of the version (anything with a `.` in it will cause this to raise).

This is mostly for my own edification as I screwed up the regex the first time around :)

rossbar:
> Not sure if post-merge review is helpful, but I left a couple of minor improvement suggestions.

Helpful indeed! I've incorporated the suggestions in #610 - thanks!

> This surprises me because I thought the root cause of https://github.com/pygraphviz/pygraphviz/pull/566 was users BYOing an uncontrolled version of Graphviz. If the only supported workflow is Graphviz libraries being bundled with the Pygraphviz installation, just drop support for Graphviz < 13 and only ever bundle Graphviz ≥ 13. What am I missing?

This was my original thought too, and why I had originally proposed #592 as an alternative to the graphviz version checking. However, after discussing with other developers, there are two main groups of people whom will still be building from source that we'd like to support:
 1. Folks on platforms where we don't provide wheels (e.g. non x86_64/aarch64 architectures)
 2. Package maintainers who are responsible for packaging pygraphviz for a specific distro

We didn't want to force the latter group (yet) to impose hard upper/lower bound pins on either graphviz/pygraphviz.

At some point in the future I think we *will* just require graphviz v13 as the minimum required version so we can drop all the switching; but we didn't want to do so for the first 2.X release in an effort to maintian support for the widest span of users.

PR #604, opened 2026-06-06
CI: Build macos wheels against macos 11, and move cibuildwheel to release pipeline

MridulS:


rossbar (approved):
This LGTM - thanks @MridulS !

I think the general plan of supporting as wide a range as possible for macos for the initial 2.0 release is the right way to go (even if it's a little slower/expensive).

PR #605, opened 2026-06-06
Graphviz support across v13 API change - no runtime check

rossbar:
Alternative to #603 - same basic pattern for API management but without the runtime version check per Smattr's [suggestion](https://github.com/pygraphviz/pygraphviz/pull/603#issuecomment-4640014832)


rossbar:
Using the existence of `agnew` as a proxy for the version didn't work out of the box - I wonder if the `agnew` macro was defined for *all* graphviz versions prior to v13 (going all the way back to v2.42, which is still picked up in ubuntu24).

We'll push ahead with #603 for now; if anyone's interested please feel free to revisit - doing away with the runtime `dot -V` checks would certainly be an improvement!

PR #608, opened 2026-06-07
CI: Build macos wheels with quartz; no gd

MridulS:


rossbar (approved):
Great - the renderings now look as expected (give or take a windows text location upstream issue)!

PR #610, opened 2026-06-07
More 2.0 cleanups

rossbar:


setup.py
```diff
@@ -15,15 +15,15 @@
...
     # reliable way to get portable version info across all versions.
     # If, in the future, it is possible/reasonable to set Graphviz 14 as a
     # minimum supported version, then this should be replaced with the macro
     # See gh-573 for further discussion
     try:  # Better exception message if graphviz not installed
         version_str = subprocess.check_output(["dot", "-V"], stderr=subprocess.STDOUT)
-    except FileNotFoundError:
-        raise FileNotFoundError(
+    except FileNotFoundError as e:
+        raise FileNotFoundError from e(
```
Smattr: Sorry, I think I was unclear. I meant `raise FileNotFoundError(…) from e`, not with the `from e` before the paren. I think the latter probably does not work or does not do what we want.

rossbar: Yup that's just me not paying attention - thanks!

PR #611, opened 2026-06-08
Update type of id param to IDTYPE in swig wrapper

### LINKED ISSUE #174: Incorrect types in swig input file
The agattr_label function is declared to return int, but it should return Agsym_t * instead.  Also, the second parameter of agidnode and the fourth parameter of agidedge should have type IDTYPE, rather than unsigned long (which may or may not be the same type as IDTYPE).

MridulS:
Fixes https://github.com/pygraphviz/pygraphviz/issues/174


rossbar (approved):
Let's give it a try!

PR #614, opened 2026-06-12
CONF: escaped slashes in windows pathspec.

rossbar:
See if this allows us to get rid of the setup.py hack

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -209,14 +209,14 @@ jobs:
             {package}/pygraphviz/bin/
           CIBW_BEFORE_BUILD_WINDOWS: >
             mkdir {package}\\pygraphviz\\bin &
-            copy C:\graphviz\bin\gc.exe {package}\pygraphviz\bin\ &
-            copy C:\graphviz\bin\acyclic.exe {package}\pygraphviz\bin\ &
-            copy C:\graphviz\bin\gvpr.exe {package}\pygraphviz\bin\ &
-            copy C:\graphviz\bin\gvcolor.exe {package}\pygraphviz\bin\ &
-            copy C:\graphviz\bin\ccomps.exe {package}\pygraphviz\bin\ &
-            copy C:\graphviz\bin\sccmap.exe {package}\pygraphviz\bin\ &
+            copy C:\\graphviz\\bin\\gc.exe {package}\\pygraphviz\\bin\\ &
+            copy C:\\graphviz\\bin\\acyclic.exe {package}\\pygraphviz\\bin\\ &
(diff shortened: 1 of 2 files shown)


rossbar:
> See if this allows us to get rid of the setup.py hack

Looks like this indeed was the issue :expressionless:

.github/workflows/release.yml
```diff
@@ -182,7 +182,7 @@ jobs:
             PATH="$PATH:/opt/graphviz/bin"
           CIBW_ENVIRONMENT_WINDOWS: >
             GRAPHVIZ_PREFIX=C:\graphviz
-            PATH="C:\graphviz\bin;$PATH"
+            PATH="C:\\graphviz\\bin;$PATH"
```
dschult, line 185: Is the escaped slash needed in the previous line? 
`GRAPHVIX_PREFIX=C:\graphviz`

rossbar: Good question - it's clearly working as-is otherwise the subsequent step that copies scripts from their install location to the specified prefix would fail. Perhaps it has to do with whether quotes are used? Either way - I'm going by results-based-coding when it comes to windows: my only real signal for whether windows syntax is valid is whether CI works!

rossbar: This was a very nice catch @dschult - it made me realize that I wasn't sure there were any tests for the CLI-based functionality that was being packaged up in the wheels. I added tests for `tred` (one of the CLI-only fns) in #615 and indeed it did catch issues that I'm 99% are related to broken path specs in the windows wheel setup.

It turns out there are multiple places where the windows path sep char needs to be escaped to work properly in the yaml config.

rossbar:
Okay - here's the deal AFAICT - it turns out the pathspec stuff for windows was incorrect universally in the CIBW configuration, but we never had any tests which hit the CLI code branches.

I added such a test in gh-615 and used it to confirm that the updated escaped-paths work for windows now. The way I've decided to organize it is as follows:
 - This PR now contains all of the necessary path updates to get the windows wheels (including graphviz CLI components) working
 - gh-615 contains *only* the added test which uncovers this behavior.

What this means is: this PR should pass on `main`, but the windows build job in gh-615 will be failing. Once this is merged, gh-615 can be rebased on `main` and it will work!

If this is too complicated, I'm happy to combine both PRs but then I'd want to rebase-and-merge to keep the CI conf and the testing separate in the history. Just LMK!

dschult (approved):
> What this means is: this PR should pass on main, but the windows build job in https://github.com/pygraphviz/pygraphviz/pull/615 will be failing. Once this is merged, https://github.com/pygraphviz/pygraphviz/pull/615 can be rebased on main and it will work!

This approach sounds good to me! But I'll let you do the merging etc so you know what is what.  I approve this PR

PR #615, opened 2026-06-13
TST: Add tests for `tred` method

rossbar:
`tred` is one of the functions that still requires the CLI of graphviz, so this doubles as a packaging test for the path-munging approach used for the pygraphviz 2.0 wheels.

Loosely inspired by @dschult 's review in gh-614 which made me realize I'm not sure an explicit test of the CLI-based functionality exists in the test suite. Prior to pygraphviz2.0 there was no distinction between "library-based" and "cli-based" functions, as graphviz was required to be installed on the user's system no matter what.

### DIFF
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -1421,13 +1421,29 @@ class AGraph:
     def tred(self, args="", copy=False):
         """Transitive reduction of graph.  Modifies existing graph.
 
-        To create a new graph use
+        See the graphviz "tred" program for details of the algorithm.
+
+        Examples
+        --------
+        tred modifies the graph in-place by default:
 
         >>> import pygraphviz as pgv
         >>> A = pgv.AGraph(directed=True)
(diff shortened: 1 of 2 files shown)


rossbar:
The expectation is the windows build will be failing here until gh-614 is merged - see https://github.com/pygraphviz/pygraphviz/pull/614#issuecomment-4697850971 for more details!

dschult (approved):
Nice! Good to have tests of the CLI tools and `tred` is a good choice.

rossbar:
Okay everything worked as expected - tests were failing before gh-614 (i.e. commit 97189597), and are now passing after merging/rebasing this PR on 923cf8e.

Will go ahead and merge now - @MridulS this may motivation for another rc, though I don't think it's absolutely necessary.

MridulS:
https://pypi.org/project/pygraphviz/2.0rc2/ is out 🚀

PR #616, opened 2026-06-14
CI: See what happens when freethreading enabled.

rossbar:
Adds 3.14t to the Python testing matrix to see what happens with freethreaded builds.

pygraphviz does not explicitly support freethreaded Python, so the expectation is python will fallback to the GIL when pygraphviz is imported (this is what happens). This PR *forces* the GIL to stay off just to see what happens.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.11", "3.12", "3.13", "3.14"]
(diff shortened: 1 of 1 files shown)


rossbar:
Well, things pass but that doesn't mean much - our test suite doesn't contain any threading examples.

IMO it doesn't hurt to enable freethreaded jobs in the testing workflows. Note that we should **not** add them to the cibuildhweel jobs without actually doing the work to make the wrapper threadsafe, as we don't want to accidentally publish freethreaded wheels.

Issue #171, opened 2018-09-06
trailing character in dot file doesn't cause parse error but breaks Agraph constructor


jcaplan:
given the file hello.gv.txt:
```
digraph G {Hello->World}"
```
with a trailing " character. Dot is still able to parse the file.
Calling:
```
graph = AGraph('./hello.gv.txt')
```
also succeeds.
However, calling Agraph again will always cause a DotError
```
graph = AGraph('./hello.gv.txt')
graph = AGraph('./profile.gv.txt')
```
results in:
```
Error: syntax error in line 1 near '"'
Traceback (most recent call last):
  File "/home/jcaplan/.local/lib/python3.5/site-packages/pygraphviz/agraph.py", line 1201, in read
    self.handle = gv.agread(fh, None)
ValueError: agread: bad input data

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test_graphviz.py", line 5, in <module>
    graph = AGraph('./profile.gv.txt')
  File "/home/jcaplan/.local/lib/python3.5/site-packages/pygraphviz/agraph.py", line 162, in __init__
    self.read(filename)
  File "/home/jcaplan/.local/lib/python3.5/site-packages/pygraphviz/agraph.py", line 1203, in read
    raise DotError
pygraphviz.agraph.DotError
```
The behaviour is odd and the error message is super confusing (especially if you don't catch the typo in the hello file). I'm guessing there is some shared character buffer somewhere which is not being cleared properly.

rossbar:
FWIW this *seems* to be a graphviz issue, or at least it was fixed by graphviz. The tricky part is knowing exactly which version of graphviz introduced the fix --- I've narrowed it down to >v12.2.1 - see gh-620 for details.

PR #620, opened 2026-06-15
TST: Add regression test for gh-171.

rossbar:
Turns the example for the bug ID'ed in gh-171 into a unit test.

I was unable to produce locally, so hopefully everything is fixed. Let's see if CI is green on all platforms.

### DIFF
--- a/pygraphviz/tests/test_readwrite.py
+++ b/pygraphviz/tests/test_readwrite.py
@@ -1,6 +1,9 @@
 import pygraphviz as pgv
 
+graphviz_major_version = int(pgv.__graphviz_version__.split(".")[0])
 
+
+@pytest.mark.skipif(graphviz_major_version < 13, reason="Graphviz version too old")
 def test_multiple_reads_same_source_trailing_character(tmp_path):
     """Ensure multiple reads from the same text file with an unexpected trailing
     character don't cause agread fails. See gh-171."""
--- a/pygraphviz/tests/test_readwrite.py
+++ b/pygraphviz/tests/test_readwrite.py
@@ -1,6 +1,21 @@
 import pygraphviz as pgv
 
 
+def test_multiple_reads_same_source_trailing_character(tmp_path):
+    """Ensure multiple reads from the same text file with an unexpected trailing
+    character don't cause agread fails. See gh-171."""
+    fpath = tmp_path / "hello.gv.txt"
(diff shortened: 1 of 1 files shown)


rossbar:
Very interesting - so it passes on some, fails on others. Based on *where* it's failing, I suspect the problem was address upstream in graphviz itself, sometime after v12.2.1:

**Fails**
- cibuildwheel jobs (currently building against [graphviz v12.2.1](https://github.com/pygraphviz/pygraphviz/blob/3074e2feadf4c795a6c586c1604af2bda0179700/.github/workflows/release.yml#L12))
- ubuntu test workflows (currently using graphviz 2.42, specifically `2.42.2-9ubuntu0.1` - see action logs)

**Doesn't fail**
- windows test workflows (graphviz v15.0.0)
- My machine (linux aarch64, graphviz v14.1.4)
- brew test workflows (graphviz v15.0.0)

This is all circumstantial of course, but it really does smell like an upstream fix. Let's see if we can't update the graphviz build target version in the cibuildwheel conf. IMV though this isn't a blocker for the 2.0 release!

rossbar:
Thanks @MridulS ! This result supports the fact that gh-171 is solved by newer versions of graphviz, as the wheel builds are all now green. Now - I failed to consider how to deal with the test workflows on ubuntu, which still use the `apt`-packaged graphviz which is way behind.

In a perfect world we could `pytest.mark.skipif` on the graphviz version, but AFAIK pgv doesn't expose the underlying graphviz version info (something we should consider adding). I'm going to let this sit til we have a chance to figure out a solution for CI - still not a blocker for release IMV!

rossbar:
Okay - I merged #620 then added a `skipif` to this test conditioned on the detected graphviz version. Everything is working as expected now. Note that I don't know *exactly* which version of graphviz fixed the bug, but we know from the previous wheel builds that it's >12.2.1. IMV it's not worth tracking down exactly - worst case scenario we're skipping this test on a narrow range of graphviz version on which it would pass.

dschult (approved):

rossbar:
I took the liberty of merging this and cutting a new release candidate with all the testing/packaging changes since the rc2. I figured it was worth another rc since we changed the packaged graphviz version from 12.2.1 -> 14.1.5; also, I wanted to familiarize myself with the release process!

PR #623, opened 2026-06-16
Bump packaged graphviz version.

rossbar:
Check whether a newer graphviz works with the wheel builds. The motivation for doing so would be e.g. closing #171 .


rossbar:
Wheel builds are all green :tada: 

I'd vote to get this in for the 2.0 release, if for no other reason than there is at least one longstanding bug it seems to solve. Before doing so though I'd vote to merge the proposed new tests (#617 through #622) first.

PR #624, opened 2026-06-17
ENH: Export version info from wrapped graphviz

rossbar:
Now that graphviz is packaged within pygraphviz itself, it would be good to provide API so that users can check the underlying graphviz version.

Inspired by #620, to which we'd like to add conditional test skips for older versions of graphviz (but there's currently no way to assess the from pygraphviz itself).

Re: implementation - I'm no swig expert so I just arbitrarily picked a pattern from the [SWIG-basics docs](https://www.swig.org/Doc3.0/SWIG.html#SWIG_nn9). Of course, this involves pre-parsing the version string into ints for major, minor and patch versions; passing them in via SWIG CLI macros at build time; then reconstructing the version string at import time. There's probably a more elegant way to do this - SWIG folks feel free to weigh in!

Another decision I made was to handle all the version comparisons manually rather than relying on a nicer interface like `packaging.version`. My reasoning was to keep the build dep footprint to an absolute minimum. If we decide that adding a build-time dependency on `packaging` is acceptable, we can clean some of this up. I'd vote to hold that discussion off for a follow-up PR though!

Finally, there's the API for the graphviz version info itself. I arbitrarily chose a dunder-pattern `__graphviz_version__` to a) match the familiar `__version__` pattern and b) prevent accidental discover of the hidden `_graphviz` module if users happen to do something like `pgv._g<tab>` while searching for graphviz version info. If others prefer a different pattern that's fine with me!

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -12,6 +12,13 @@ Import PyGraphviz with
 
 >>> import pygraphviz as pgv
 
+PyGraphviz wraps `graphviz <https://graphviz.org>`_ providing a Python interface
+to graphviz's functionality.
+Information about the version of graphviz that is wrapped by pygraphviz can be
+found with
+
+>>> pgv.__graphviz_version__
+
 Graphs
 ------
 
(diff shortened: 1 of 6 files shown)


dschult (approved):
This seems quite useful -- and looks good to me. :)

rossbar:
I'm going to get this in to un-stick #620 - we can always go back and cleanup the SWIG exposure pattern in a follow-up!

PR #625, opened 2026-06-18
Add Python 3.10 support for 2.0 release

MridulS:


rossbar:
Gah - the category keyword for `catch_warnings` was added in 3.11 "/. Hopefully there's a simple testing pattern that works across all Python versions...

pygraphviz/tests/test_graph.py
```diff
@@ -43,11 +43,12 @@ def test_tred_cycle():
     """RuntimeWarning raised when cycle detected."""
     A = pgv.AGraph(directed=True)
     A.add_edges_from([(0, 1), (1, 2), (2, 0)])  # Directed 3-cycle
-    with warnings.catch_warnings(record=True, category=RuntimeWarning) as rec:
+    with warnings.catch_warnings(record=True) as rec:
```
dschult, line 46: Do you want a comment line to remember that this can go away after support for 3.10 is dropped?
OK either way for me.

MridulS: I think we can just revert this PR, once the release is out :)

rossbar (approved):
I tried the wheels from the artifact upload in an old 22.04 container (which has Python3.10 by default) and it worked as expected. +1 for adding this just for the 2.0 release!

Should we do (yet another) rc that includes the 3.10 wheels?

MridulS:
> Should we do (yet another) rc that includes the 3.10 wheels?

Let's do it! But I want to add the graphviz license in the wheel too before that!

PR #629, opened 2026-06-20
DOC: Make tutorial executable, run as part of doc build

rossbar:
This PR proposes to incorporate the `myst-nb` extension, reformat the tutorial as a [markdown-based jupyter notebook](https://myst-nb.readthedocs.io/en/v0.9.1/use/markdown.html), and ensure that the tutorial is executed at doc-build time.

This essentially a pygraphviz-version of networkx/networkx#7237 (though simpler, since there was no existing execution engine to replace).

The primary benefit here is ensuring that the tutorial documentation is actually correct - activating the execution caught [one cell that actually raises an exception due to invalid dot format](https://github.com/pygraphviz/pygraphviz/blob/d8d49391b0109c385ffb413d86c8989c5957d4d1/doc/source/tutorial.rst?plain=1#L42). It also adds another layer of integration testing to catch exceptions/warnings/deprecations from e.g. changes in graphviz.

Finally - there is a relatively simple pattern to adopt to actually show the drawing results in the tutorial, which is a nice little enhancement IMO.

I used `rst2myst` to convert the tutorial to MyST format; unfortunately that doesn't do much to preserve diffs. I think there's a way I could improve this with some manually patch munging - LMK if you'd like me to try!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ doc/build
 doc/source/static/examples/*
 doc/source/reference/generated/*
 doc/source/examples/*
+doc/source/file.*
 
 # Generated when 'python setup_egg.py'
 pygraphviz.egg-info/
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -68,6 +68,12 @@ add_module_names = False
 pygments_style = "sphinx"
 
 
+def setup(app):
+    # Workaround to prevent duplicate file warnings from sphinx due to having
+    # both myst-nb and sphinx-gallery generating executable files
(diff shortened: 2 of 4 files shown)


rossbar:
Here's what the tutorial looks like on [main](https://pygraphviz.github.io/documentation/latest/tutorial.html) | [this PR](https://output.circle-artifacts.com/output/job/1767f2b2-b99d-4ff4-b2a8-6c9a431e9ebe/artifacts/0/doc/build/html/tutorial.html)

PR #638, opened 2026-07-08
BUG: Use "nop2" instead of "neato -n2" for rendering with pre-computed positions

### LINKED ISSUE #637: Regression on dot.draw if submitted dot has pre-computed positions with subgraphs
Subgraph containment ignored on the resulant svg/png/pdf renders

Expected outcome

Image

Actual outcome

Image

Minimal reproduction steps

```from pygraphviz import AGraph
from pathlib import Path
from graphviz import Source

import subprocess

dot_string="""
digraph G {
    graph [compound=true];

    node [shape=box];

    subgraph cluster_A {
        label="A";
        a1 [pos="50,100!"];
    }

    subgraph cluster_B {
        label="B";
        b1 [pos="200,100!"];
    }

    a1 -> b1 [lhead=cluster_B, ltail=cluster_A, dir=both];

    graph [bb="0,0,300,200"];
}

"""
def render_bytes(dot_string_value: str, fmt="svg") -> bytes:
    gv_source = Source(dot_string)
    return gv_source.pipe(format=fmt)


outfile_directory=Path("/tmp")
sample_dot_file = outfile_directory / "sample.dot"
sample_dot_file.write_text(dot_string)
dot = AGraph(string=dot_string)
dot.layout(prog="dot")
src = Source(dot_string)


dot.write(outfile_directory / "after_layout_sample.dot")
dot.draw(outfile_directory / "overlay.dot", format="dot")
dot.draw(outfile_directory / "overlay.svg", format="svg")
dot.draw(outfile_directory / "overlay.pdf", format="pdf")
dot.draw(outfile_directory / "overlay.png", format="png")

src.render(outfile_directory / "gv", format="svg", cleanup=True)

bytes_render = Path(outfile_directory / "gv_bytes.svg")
bytes_render.write_bytes(render_bytes(dot_string))

subprocess.run(["dot", "-Kdot", "-Tsvg", sample_dot_file, "-o", outfile_directory / "cli.svg"])
subprocess.run(["dot", "-Kdot", "-Tpng", sample_dot_file, "-o", outfile_directory / "cli.png"])
subprocess.run(["dot", "-Kdot", "-Tpdf", sample_dot_file, "-o", outfile_directory / "cli.pdf"])

```

Definite regression behaviour from 1.14 - have worked around by falling back to graphviz direct for bytes rendering - but something is up!

Oh, and a thousand thanks for releasing 2.0 - was kinda giving up hope after 18 months of hacking dynamic links and playing install roulette :)

(NB: Just created a simpler minimal example)

rossbar:
Closes #637 

While investigating that issue, I stumbled across [this graphviz discourse post](https://forum.graphviz.org/t/what-is-the-purpose-of-the-n2-option-with-neato/1107) which I think highlights where the issue may lie, and what to do about it[^1].

From my reading - the intended purpose of `neato -n2` is to trigger the rendering engine for drawing *without* modifying the node/edge layout at all. This appeared to work okay with the `neato` CLI, which is what was called in pygraphviz<2.0. As of 2.0 however, we made the switch to rendering via the library interface - and there appear to be subtle changes with how well/whether the `-n2` flag is followed. This is highlighted not only by #637, but I also noticed that the Knuth miles example in the gallery is rotated 180 degrees compared to it's original (i.e. pygraphviz<2.0) orientation.

The fix for both cases is to replace the `"neato -n2"` incantation with `"nop2"`, which AIUI is a sentinel "layout" primarily designed *only* to trigger the rendering engine without messing with node/edge layouts at all.

FWIW I do not think this behavior was/is reliably tested, but I did add the example from #637 to the gallery to at least visually confirm the expected outcome. I'd like to add `pytest-mpl` for more rigorous testing for rendering regressions, but will do so in a follow-up PR when I have time!

[^1]: The last response in that thread (from Apr 2022) provides background for why things are the way they are.

### DIFF
--- a/examples/plot_miles.py
+++ b/examples/plot_miles.py
@@ -93,5 +93,5 @@ if __name__ == "__main__":
 
     G.write("miles.dot")
     print("Wrote miles.dot")
-    G.draw("miles.png", prog="neato", args="-n2")
+    G.draw("miles.png", prog="nop2")
     print("Wrote miles.png")
--- a/examples/plot_subgraph_boundaries.py
+++ b/examples/plot_subgraph_boundaries.py
@@ -1,0 +1,36 @@
+"""
+Subgraph with Boundaries
+========================
+
+Draw a subgraph with bounding boxes demarcating both the nodes and the subgraphs
+"""
+
+import pygraphviz as pgv
+
+# A graph in `.dot` format specifying two nodes, each within their own subgraph,
+# and a bi-directional edge between them
+dot_str = """
+digraph G {
+    graph [compound=true];
+
+    node [shape=box];
+
+    subgraph cluster_A {
+        label="A";
+        a1 [pos="50,100!"];
+    }
+
+    subgraph cluster_B {
+        label="B";
+        b1 [pos="200,100!"];
+    }
+
+    a1 -> b1 [lhead=cluster_B, ltail=cluster_A, dir=both];
+
+    graph [bb="0,0,300,200"];
+}
+"""
+
+A = pgv.AGraph(dot_str)
+A.layout("dot")
+A.draw("subgraph_bdries.png")
--- a/pygraphviz/agraph.py
+++ b/pygraphviz/agraph.py
@@ -1580,8 +1580,7 @@ class AGraph:
 
         if prog is None:
             if self.has_layout:
-                prog = "neato"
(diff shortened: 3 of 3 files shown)


dschult (approved):
Good Sleuthing!

Looks good to me.

Issue #630, opened 2026-06-26
Use pkg-config to obtain include_dirs and library_dirs


dotlambda:
While there is `GRAPHVIZ_PREFIX` it's possible for the library and the headers to be installed under different prefixes. Using pkg-config would allow finding the correct paths on any system on which it is available.

rossbar:
I assume you mean for the install-from-source code path [here](https://github.com/pygraphviz/pygraphviz/blob/d12524bcd379890971b01e1a82d9cc73f28b5a90/setup.py#L61-L81)? The primary use of `graphviz_prefix` is as a proxy to determine whether or not it's a cibuildwheel run or not. In the case where it's not, I'd be +1 on switching to a more flexible discovery recipe - PRs welcome for patterns that alleviate packaging issues on you platform!

chenrui333:
Homebrew/homebrew-core#292469 exposed a related source-install failure with pygraphviz 2.0 on Linuxbrew, where Graphviz is installed under a nonstandard prefix.

Without `GRAPHVIZ_PREFIX`, the build cannot find Graphviz's plugin libraries. Setting `GRAPHVIZ_PREFIX` allows the extension to link, but that branch intentionally omits `runtime_library_dirs` for wheel repair. A source installation therefore builds successfully and then fails on import with:

```text
ImportError: libgvplugin_core.so.8: cannot open shared object file: No such file or directory
```

The extension also directly needs `libgvplugin_{core,dot_layout,neato_layout,gd,pango}`, which live under `<libdir>/graphviz`.

I plan to open a PR that uses `pkg-config` (`libgvc`) for non-wheel source-build discovery, derives the plugin directory from its reported `libdir`, and supplies both directories as runtime library paths. The existing `GRAPHVIZ_PREFIX` path can remain dedicated to cibuildwheel builds, with the current hard-coded discovery paths retained as a fallback when `pkg-config` is unavailable.

Reference: https://github.com/Homebrew/homebrew-core/pull/292469

PR #640, opened 2026-07-11
DOC: Document custom Graphviz build paths

chenrui333:
Summary:
- document the `build_ext` options for Graphviz installations in nonstandard prefixes
- include Graphviz's plugin directory in link and runtime paths

Refs #630.

References:
- Homebrew/homebrew-core#292469


rossbar (commented):
At face value this seems to add more complexity than it actually helps. I'd lean towards simply recommending the `--global-option` flag to explicitly specify necessary paths. This is the same recipe for building from source for pygraphviz 1.X when headers/libs were not on the default discovery paths

chenrui333:
That makes sense. I've simplified this to a documentation-only change using the existing `build_ext` options. For pygraphviz 2.0, the recipe includes Graphviz's plugin directory in both `-L` and `-R` so source installs can link and import successfully. I've dropped the discovery code/tests and left #630 open for any future automatic-discovery work.

rossbar (commented):
IMO the existing documentation is specific. I'm -1 on adding system-specific instructions: presumably, folks who are installing from source are aware of the constraints on their system. For example, recommending runtime library paths is bad advice for folks building on windows...

Ultimately, building from source really just depends on the system - so there has to be an expectation that the folks who are doing so are aware of those issues. I'd just as soon leave these docs as-is.

PR #644, opened 2026-08-03
MAINT: simplify compressed filename handling

dschult:
This PR switches `_get_fh` from manually checking for compressed files (using gzcat or bz2) to using the Python standrad library tools which do the same thing. In this case:  
`fh = fileinput.hook_compressed(path, mode=mode)`
gives us the file handle for the open file and decompresses the contents if the path ends with a compressed suffix like `.gz`.

This also reduces our exposure to user-input tricks.


rossbar (approved):
+1 for replacing any `os.popen` s with stdlib functionality! Thanks @dschult for finding this very nice solution.

